PR #216, opened 2025-06-06
copyToClipboard

### LINKED ISSUE #210: copyToClipboard
allow scenes to copy text to clipboard

- method in restrictedActions
- allow after user action only
- add permission management
- show popup

robtfm:
- add `copyToClipboard` to restrictedActions.js
  - active scene only
  - shortly after user interaction only
  - with related permission and notification on use

closes #210

note: doesn't work for web currently

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9101,6 +9101,7 @@ dependencies = [
  "common",
  "comms",
  "console",
+ "copypasta",
  "dcl_component",
  "ethers-core",
  "ipfs",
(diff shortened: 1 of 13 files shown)


kuruk-mm (approved):
Great. I compared [arboard](https://crates.io/crates/arboard) crate and copypasta crate.

The advantage for arboard is backed by 1Password, supports copying images, and it has better support for Wayland.

But copypasta looks simpler, rock solid.

robtfm:
i will fork copypasta to add wasm, the browser api is async so it's a pretty fundamental change, unlikely to be accepted as a pr (to arboard or copypasta)

PR #220, opened 2025-06-12
feat: livekit for web

kuruk-mm:


deploy/web/index.html
```diff
@@ -122,6 +122,7 @@ <h1>Bevy Explorer</h1>
         </div>
     <canvas id="mygame-canvas" width="1280" height="720"></canvas> 
     
+    <script src="https://cdn.jsdelivr.net/npm/livekit-client/dist/livekit-client.umd.min.js"></script>
```
kuruk-mm, line 125: @robtfm please check this
this is how I'm adding LiveKit JS SDK to the project. I could make it fetch on runtime, and only if I'm with the livekit feature + wasm...

Let me know what you think.

leanmendoza (commented):
:clap:

crates/av/src/microphone.rs
```diff
@@ -1,19 +1,12 @@
 use bevy::prelude::*;
```
robtfm, line 1: i think we could move the microphone.rs code into livekit_native.rs, so it's parallel with the web version (we can do in a separate pr if you prefer)

crates/comms/src/livekit_room.rs
```diff
@@ -104,24 +100,36 @@ fn connect_livekit(
...
+            });
+        }
 
-        std::thread::spawn(move || {
-            livekit_handler(transport_id, remote_address, receiver, sender, subscription)
-        });
+        #[cfg(target_arch = "wasm32")]
+        {
+            // For WASM, we directly call the handler which will spawn the async task
+            let receiver = Arc::new(Mutex::new(receiver));
```
robtfm, line 122: nit: you don't need the arc<mutex> here, we have it for native because we use a retry loop so we needed to clone it

crates/comms/src/livekit_web.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+async fn run_livekit_session(
+    transport_id: Entity,
+    address: &str,
+    token: &str,
+    app_rx: Arc<Mutex<Receiver<NetworkMessage>>>,
+    sender: Sender<PlayerUpdate>,
+) -> Result<(), anyhow::Error> {
+    let room = connect_room(address, token)
```
robtfm, line 138: can we please wrap this in a retry loop, similar to native in `livekit_handler`. the native version repeatedly connects until the sender is closed (i.e when we change island / scene).

crates/comms/src/livekit_web.rs
```diff
@@ -0,0 +1,279 @@
...
+    close_room(&room)
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to close room: {:?}", e))?;
+
+    Ok(())
+}
+
+async fn handle_room_event(event: JsValue, transport_id: Entity, sender: Sender<PlayerUpdate>) {
+    // Parse the event from JavaScript
+    // This is a simplified version - you'll need to properly parse the JavaScript event object
```
robtfm, line 198: all this should be handled by serde_wasm_bindgen - you can make a rust struct with `#[derive(serde::Deserialize)]` and `#[serde(rename_all='camelCase')]` and then deserialize a JsValue into it with `serde_wasm_bindgen::from_value(js_value)`, we do this in dcl_wasm already.

crates/comms/src/livekit_web.rs
```diff
@@ -0,0 +1,279 @@
...
+                        if let Ok(participant) =
+                            js_sys::Reflect::get(&event, &JsValue::from_str("participant"))
+                        {
+                            if let Ok(identity) =
+                                js_sys::Reflect::get(&participant, &JsValue::from_str("identity"))
+                            {
+                                if let Some(identity_str) = identity.as_string() {
+                                    if let Some(address) = identity_str.as_h160() {
+                                        // Convert payload to bytes
+                                        let data = js_sys::Uint8Array::new(&payload);
```
robtfm, line 214: not sure how serde will handle this, but it might work out of the box using Vec<u8> in the struct

crates/comms/livekit_web_bindings.js
```diff
@@ -0,0 +1,180 @@
...
+            payload: payload,
+            participant: participant,
+        });
+    });
+    
+    room.on(LivekitClient.RoomEvent.TrackSubscribed, (track, publication, participant) => {
+        // For audio tracks, automatically play them
+        if (track.kind === 'audio') {
+            const audioElement = track.attach();
+            audioElement.play().catch(e => console.warn('Failed to play audio:', e));
```
robtfm, line 132: this means we lose the spatial position for the received audio...

it seems a shame to send audio data back to rust and re-process it though ... we could possibly keep the processing in js and just send the volume / panning info from rust to js (if audioElement allows us to manage the left/right channel volumes separately).

this can be a follow up though, maybe to consider alongside threading the audio better for web - we may want all audio processing to be browser-side.

crates/system_ui/src/mic.rs
```diff
@@ -29,6 +28,8 @@ impl Plugin for MicUiPlugin {
         app.add_systems(Startup, setup.in_set(SetupSets::Main));
         app.add_systems(Update, update_mic_ui);
 
+        app.init_resource::<MicState>();
```
robtfm, line 31: let's move this to the comms plugin please, eventually this system ui module will probably get removed entirely

deploy/web/index.html
robtfm: as discussed, let's vendor it locally. don't worry about runtime fetching, it's only 100k which is like 0.2% of the wasm and i don't think we will ever want the feature disabled in releases anyway.

robtfm (commented):
looking good. couple of small reshuffles and i would like to switch the js-value decoding to serde_wasm_bindgen before merging, but i think that's my only blockers

crates/comms/livekit_web_bindings.js
robtfm: yes that works. don't worry about setting the values yet, we can do that later.

PR #224, opened 2025-06-23
Bevy 16

robtfm:


kuruk-mm (approved):
It is working GREAT! The only requested change it is to change

```
let version = "webgpu proof of concept".to_string();
```

Issue #203, opened 2025-06-06
Wasm thread blocking behaviour


robtfm:
exact cause unclear, but it seems like sometimes new scenes are blocked from spawning until an existing heavy scene is despawned. may be due to scene runner's jobs-in-flight limit, and the scene worker initial rpc call post waiting for a response?
observed particularly around ~135,135 walking from 150,150

PR #225, opened 2025-07-01
webgpu fixes

### LINKED ISSUE #199: separate thread for asset processing
asset processing (particularly gltfs and resampling of 16bit channel images) runs on the bevy engine main thread. this causes significant hitching / blocking. investigate using a separate thread for asset processing

### LINKED ISSUE #201: gltf panic crashes engine in wasm
invalid gltf data causes the engine thread to crash in wasm as panic_unwind / catch_unwind are not supported. fix this particular panic at the root to avoid the issue as best we can

### LINKED ISSUE #214: thread logging causes overflow after 128 scenes are loaded
- can be worked around by not adding the `EnteredSpan` in dcl::op_crdt_recv_from_renderer.
- is not solved by `wasmContext.free()` or calling to a `op_clean_up(_: WorkerContext)` wasm function.
points to a deeper issue with resource dropping which needs investigation

robtfm:
updates for webgpu
- use separate worker for asset loading
- auto-resize large images
- increase wasm thread stack size to 10mb
- feature-gate scene workers trace span due to hardcoded 128 thread max limit
- embed scene import scripts in wasm module
- use spin::RwLock for resources that can be contended (wallet, ipfs collection metadata)
~~- use blob for scene worker script~~ - disabled as this fails with livekit feature
- expose RenderAssetBytesPerFrame to js api, default to 1e6
- reduce initial wasm page-pool (helps with oom, not clear why)
- remove SystemApi module from non-superuser scenes
- clean wasm thread context properly on clean worker exit
- reduce max-concurrent downloads to 8

fixes #214
fixes #201
fixes #199


robtfm:
fixes #214
fixes #203
fixes #201
fixes #199

PR #253, opened 2025-07-29
Web video

### LINKED ISSUE #252: builtin ui text flickers on web
loading genesis plaza (for example), the loading panel text sometimes disappears for a few frames.

possibly due to fontmap partial uploads (when a new character is used, it causes the fontmap texture to be re-uploaded, if this doesn't complete within the frame the texture is then unavailable for all text).

seems to happen only on web

robtfm:
- support `PbVideoPlayer` on web
- support `PbAudioPlayer` on web
- fix #252 by allowing textures to force override the gpu bytes-per-frame limit, set fontmap texture to force override

implementation:
- use `HtmlVideoElement` and `HtmlAudioElement` to source the data
- copy video textures in gpu queue using webgpu's `copy_external_image_to_texture`
- don't use audio sink, use the embedded audio of the html element directly
- report events using rust-managed callbacks
- use Hls.js to support m3u8 playlists

### DIFF
--- a/crates/av/src/html_video_player.rs
+++ b/crates/av/src/html_video_player.rs
@@ -142,6 +142,8 @@ pub struct HtmlMediaEntity {
     new_frame_time: Arc<AtomicU32>,
     state: Arc<Mutex<VideoState>>,
     _closures: Vec<Closure<dyn FnMut()>>,
+    frame_closure: Rc<RefCell<Option<Closure<dyn FnMut(f64, JsValue)>>>>,
+    frame_callback_handle: Rc<RefCell<Option<u32>>>,
 }
 
 /// safety: engine is single threaded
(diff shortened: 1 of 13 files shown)


robtfm:
@kuruk-mm would appreciate your opinion on embedding hls

crates/av/src/html_video_player.rs
```diff
@@ -0,0 +1,623 @@
...
+            wgpu::CopyExternalImageDestInfo {
+                texture: &gpu_image.texture,
+                mip_level: 0,
+                origin: wgpu::Origin3d::ZERO,
+                aspect: wgpu::TextureAspect::All,
+                premultiplied_alpha: false, // Video frames are not typically premultiplied.
+                color_space: wgpu::PredefinedColorSpace::Srgb,
+            },
+            request.size,
+        );
```
kuruk-mm, line 621: Amazing

kuruk-mm (approved):
I'm totally impressed! Great job! It works very smoothly, even smoother than on desktop.

I also love how you're copying the texture into memory to avoid CPU usage 💪 

About embedding HLS, I don't see any problem. Looks great.

PR #254, opened 2025-08-05
Lights reconciliation

### LINKED ISSUE #250: reconcile light message types
we've discussed and reconciled differences in our light implementation with foundation, this requires some changes for us. 

this should be done when the lights protocol messages are merged into main.

robtfm:
closes #250 

- use agreed definition for `PbLightSource`
  - spotlight vs pointlight as `type`
  - spotlight angles in degrees instead of radians
  - default intensity 16k instead of 10k..
  - support shadow mask texture
  - interpret negative range as unbounded
  - TODO: webgpu support for light textures requires bindless
  
bugfixes
- avoid crash on recreating text entry field while active
- avoid flickering shadows when lights are updated every frame

https://github.com/user-attachments/assets/a99cb365-4d76-48d5-813f-7807e5739815


robtfm:
requires protocol update (tbd)
requires update to example scene

PR #259, opened 2025-08-09
cache shaders and ipfs files on web

robtfm:
improve startup times (after the first time) by storing details of required shaders, and compiling them at startup in parallel with wasm lib loading. 
also use a service worker to cache IPFS content requests: the browser automatically caches fetches, but doesn't have the context to understand that `peer-eu1.decentraland.org/content/bafkreixyz` is the same as `peer-eu2.decentraland.org/content/bafkreixyz`.

shader caching takes my "fully playable time" from 41 seconds to 27 seconds, and avoids the "only hands and head" issue.
content file caching benefit depends on network speed.

todo: see if we can pre-populate the cache as well, i think this depends on platform and graphics card (and maybe browser version) though. we would need some steam shader-cache style approach where users submit their spec and requirements to a server, which then provides it to other users.

also todo: see if we can get some feedback about gpu uploads so we can only drop the player into the world when the assets are actually ready to render (avoid the "empty world" issue which is still there / worse).

https://github.com/user-attachments/assets/8ef15a38-db01-455d-92c5-16d24c1d3ab9

https://github.com/user-attachments/assets/70f87376-233d-45fe-aad4-915739320d42

### DIFF
--- a/crates/ipfs/src/lib.rs
+++ b/crates/ipfs/src/lib.rs
@@ -1212,7 +1212,7 @@ impl AssetReader for IpfsIo {
 
                 // in wasm we add a custom header to allow the service worker to cache ipfs requests across content servers
                 #[cfg(target_arch = "wasm32")]
-                let request = if ipfs_path.content_path().is_some() {
+                let request = if ipfs_path.content_path().is_some() && ipfs_path.should_cache() {
                     request.header("X-IPFS", "true")
                 } else {
                     request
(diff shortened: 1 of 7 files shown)


kuruk-mm (approved):
image

Looks like it is caching even if you change the realm

Regarding the shader cache time, I don't see any differences in time, but I stopped seeing only hands and heads.

PR #264, opened 2025-08-13
Browser pointer lock

robtfm:
fix browser pointer locking behaviour. when we lock the cursor, chrome shows a popup saying "press escape to unlock cursor". if that is pressed, chrome consumes the escape key press and unlocks the cursor, but the app is still hiding the cursor and acting as though it is locked, which results in weird state.

solution: in web, check the cursor state via web-sys api and react appropriately if the cursor has been force unlocked by the browser

there is still a bit of weirdness if you
- lock with right click
- press escape to unlock
- right-click to try and lock again within about 2 seconds
because the browser won't allow you to lock the cursor again for a few seconds after it is force-unlocked. but i think this is good enough

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8545,6 +8545,7 @@ dependencies = [
  "spin 0.9.8",
  "tokio",
  "tungstenite 0.21.0",
+ "wasm-bindgen",
  "wasm-bindgen-futures",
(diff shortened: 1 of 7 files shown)


kuruk-mm (approved):
tested and it is working great

PR #265, opened 2025-08-13
use physical coords for cursor ui position

robtfm:
bevy 0.16 switched from logical to physical coords for ui position, which meant my patch (for diegetic ui) was inconsistent. fix it


kuruk-mm (approved):
Working great!

PR #275, opened 2025-08-29
Image border radius

robtfm:
constrain images to border radius corners when specified


robtfm:
closing in favour of #279

PR #289, opened 2025-09-30
Fix wasm frame limits

robtfm:
- fix wasm throttling / fps limits

### DIFF
--- a/crates/imposters/src/render.rs
+++ b/crates/imposters/src/render.rs
@@ -489,8 +489,8 @@ pub struct ImposterSpecManager<'w, 's> {
     plugin: Res<'w, DclImposterPlugin>,
 }
 
-const MAX_ASSET_LOADS: usize = 40;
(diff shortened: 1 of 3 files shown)


crates/imposters/src/render.rs
```diff
@@ -489,8 +489,8 @@ pub struct ImposterSpecManager<'w, 's> {
     plugin: Res<'w, DclImposterPlugin>,
 }
 
-const MAX_ASSET_LOADS: usize = 40;
-const MAX_ASSET_DOWNLOADS: usize = 10;
+const MAX_ASSET_LOADS: usize = 5;
+const MAX_ASSET_DOWNLOADS: usize = 5;
```
kuruk-mm, line 493: Can we change the values based on the Platform? (Desktop & Web)

robtfm: i'm going to revert this. i thought it was causing an issue but i found the real cause

PR #373, opened 2025-11-12
Livekit web multiroom fix

robtfm:
livekit web:
- create a local track per room to fix error when unsubscribing from more than 1 room
- subscribe on room creation if mic active


robtfm:
included in #381

PR #398, opened 2025-11-21
Consume streamer audio and video tracks 

hukasu:


crates/av/src/video_player.rs
```diff
@@ -105,9 +106,26 @@ fn play_videos(
                     sink.rate = Some(rate);
                 }
                 Ok(VideoData::Frame(frame, time)) => {
-                    last_frame_received = Some(frame);
+                    last_frame_received = Some(frame.data(0).to_vec());
```
robtfm: can we please make `last_frame_received` an enum to avoid copying the bytes of every frame (even if we don't display it) - i think frames may arrive in large batches if there's a network delay

crates/av/src/video_player.rs
```diff
@@ -201,20 +220,66 @@ pub fn update_video_players(
...
-                player.source.volume.unwrap_or(1.0),
-                false,
-                player.source.r#loop.unwrap_or(false),
-            );
-            debug!(
-                "spawned av thread for scene @ {} (playing={})",
-                context.base,
-                player.source.playing.unwrap_or(true)
-            );
+            let (video_sink, audio_sink) = if player.source.src.starts_with("https://") {
```
robtfm: same here - use the video branch by default

crates/av/src/video_player.rs
robtfm: no it's fine

crates/comms/src/global_crdt.rs
```diff
@@ -183,9 +185,20 @@ pub struct ForeignPlayer {
...
+    StreamerSubscribe(
+        mpsc::Sender<StreamingSoundData<AudioDecoderError>>,
+        mpsc::Sender<LivekitVideoFrame>,
+    ),
+    #[cfg(not(target_arch = "wasm32"))]
+    StreamerUnsubscribe,
+    #[cfg(target_arch = "wasm32")]
+    StreamerSubscribe,
+    #[cfg(target_arch = "wasm32")]
+    StreamerUnsubscribe,
```
robtfm:
```suggestion
    #[cfg(target_arch = "wasm32")]
    StreamerSubscribe,
    StreamerUnsubscribe,
```

crates/av/src/html_video_player.rs
```diff
@@ -675,3 +856,55 @@ fn perform_video_copies(
...
     }
 }
+
+fn unsubscribe_to_streamer(
+    trigger: Trigger<OnRemove, ShouldBePlaying>,
+    av_players: Query<&AVPlayer>,
+    mut scene_rooms: Query<&mut Transport, With<SceneRoom>>,
+) {
+    let Ok(av_player) = av_players.get(trigger.target()) else {
+        unreachable!("ShouldBePlaying should only be present on a AVPlayer.");
```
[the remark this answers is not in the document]
robtfm: oh i see this runs before any components are removed when the entity is deleted. this is really nice, a lot of code would benefit from using this.

robtfm (approved):

PR #427, opened 2025-12-04
simpler depenetration

### LINKED ISSUE #420: spawn locations can leave player stuck in scenery
we need to check the collisions at the spawn location and adjust if required

robtfm:
- check all colliders in active scenes every frame, even if unchanged
- if player is overlapping a collider, move backwards (towards last direction) and up until no longer overlapping

closes #420


robtfm:
todo: check web perf

PR #437, opened 2025-12-11
ECS-fy comms

hukasu:


crates/comms/src/livekit/web/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+            Some("audio") => Self::Audio,
+            Some("video") => Self::Video,
+            Some(other) => {
+                error!("TrackKind was not a known kind. Was '{other}'. Assuming Audio.");
+                Self::Audio
+            }
+            None => {
+                error!("TrackKind was not a string. Assuming Audio.");
+                Self::Audio
+            }
```
hukasu: Should this be a hard error?

PR #482, opened 2025-12-22
Prefer `bevy_log`

hukasu:
Using `console_log` without `DefaultPlugins.disable::<LogPlugin>()` leads to an error being logged on the the browsers console.

Bevy's `LogPlugin` and `console_log` have the same behavior on wasm but `LogPlugin` is more flexible as it allows passing `EnvFilter`.

Also add a way to pass a `EnvFilter` during compile time


hukasu:
passing the `EnvFilter` just requires you to have `RUST_LOG` on the environment when calling `wasm-pack`, either add it to the global env of the system, or do `RUST_LOG="warn,comms::livekit=debug" wasm-pack build ...`

PR #489, opened 2025-12-30
allow non-player updates over livekit

robtfm:
authoritative server replies over livekit with participant identity "authoritative-server". we currently discard messages from participants with non-H160 identities. 
- add NonPlayerUpdate(rfc4::Message) to channel
- send for non-H160 identities
- push to scene message bus

### DIFF
--- a/crates/comms/src/livekit_native.rs
+++ b/crates/comms/src/livekit_native.rs
@@ -23,7 +23,8 @@ use dcl_component::proto_components::kernel::comms::rfc4;
 
 use crate::{
     global_crdt::{
-        GlobalCrdtState, LocalAudioFrame, LocalAudioSource, PlayerMessage, PlayerUpdate,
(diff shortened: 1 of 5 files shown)


hukasu (approved):

crates/comms/src/livekit_native.rs
```diff
@@ -393,26 +394,38 @@ fn livekit_handler_inner(
...
+                                    }
+                                    None => {
+                                        if let Err(e) = sender.send(NonPlayerUpdate {
+                                            transport_id,
+                                            message,
+                                            address: participant.identity().0.clone()
+                                        }.into()).await {
+                                            warn!("app pipe broken ({e}), existing loop");
+                                            break 'stream;
+                                        }
```
hukasu:
```rust
if let Some(address) = participant.identity().as_str().as_h160() {
}
```
only increases the indentation by 1 compared to matchs 2

PR #491, opened 2025-12-31
Stop leaking `broadcast::Receiver`

hukasu:
A `tokio::sync::broadcast::Sender` does not get closed because its `Receiver`s are dropped, so no need to leak it


robtfm:
there's another the same 10 lines up, let's remove that too pls

PR #495, opened 2026-01-02
fix textshape linear rgba

robtfm:
remove accidental Srgb(2.0,2.0,2.0) base color (probably a bad way to do linear/srgba conversion before using proper types)

before
Screenshot 2026-01-02 at 14 31 41

after
Screenshot 2026-01-02 at 14 35 59

foundation
screenshot_2026-01-02_143624_720

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3177,7 +3177,7 @@ dependencies = [
  "cssparser-macros",
  "dtoa-short",
  "itoa",
- "phf",
+ "phf 0.11.3",
  "smallvec",
 ]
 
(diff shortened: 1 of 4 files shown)


hukasu (commented):
This fixed the date being blown out, but now emojis are super dark, is that fine?

robtfm:
> This fixed the date being blown out, but now emojis are super dark, is that fine?

no ... but it wasn't great before either

new:
Screenshot 2026-01-05 at 15 45 41

old:
Screenshot 2026-01-05 at 15 48 56

foundation:
image (1)

PR #564, opened 2026-01-29
Add Taplo job

hukasu:


hukasu:
checks formatting on `toml` files

PR #603, opened 2026-02-13
Skybox time

hukasu:
Works but i didn't test all edge cases

### DIFF
--- a/crates/visuals/src/day_night.rs
+++ b/crates/visuals/src/day_night.rs
@@ -227,12 +227,27 @@ fn push_time_of_day_from_time_skip(
 
 #[expect(clippy::type_complexity, reason = "Queries are complex")]
 fn push_time_of_day_from_running_clock(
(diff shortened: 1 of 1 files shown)


robtfm:
now would be a good time to implement `getDecentralandTime` as well.

this could be done either by 
- passing the dcl time to the scene on each invocation by adding it to `RendererResponse`
- broadcasting the time over the `GlobalCrdtState`s int_sender (currently only accepts `Vec<u8>` serialized crdt messages but could be modified to accept other things)
- adding an `RpcCall` variant and a system to listen and reply

i would be inclined to go via the broadcast route, because 
- it minimises the main thread work
- the api can then return immediately
- we need to add more context to these messages to fix an outstanding bug - foreign avatar transforms are broadcast to all scenes (in `process_transport_updates`), but are not adjusted relative to the scene base translation

let's make this a follow up though, and get this one merged first

PR #632, opened 2026-02-27
fix triggerarea events sometimes not firing

### LINKED ISSUE #602: missing some trigger area exit events
https://github.com/user-attachments/assets/52227af9-7f37-4c8a-86c0-22016d02cfd3

robtfm:
fixes #602 

similar to #617 trigger timestamps need to be monotonic even for simultaneous events


hukasu (approved):
couldn't reproduce it with this fix

PR #635, opened 2026-03-02
dont panic on lagged global crdt

robtfm:


robtfm:
we create the global_crdt receiver at the same time as initializing the crdt store (as the initial state of the receiver should contain the messages after the initial crdt is taken).
but this occurs before javascript processing which can be slow, resulting in a lagged global crdt receiver.
in general this only contains semi-transient data (player positions, world time), so if it is lagged we just log an error and continue instead of shutting down

PR #637, opened 2026-03-02
update to rapier 0.31

robtfm:


regenesis-claw (changes requested):
## Rapier 0.31 Update Review

This PR updates the physics engine from Rapier 0.30 to 0.31. The refactoring is comprehensive and includes significant API changes. I found one critical bug that will cause panics.

### Critical Issue

**Bug in `get_ground()` method** (line 434-470)
- This method calls `query_pipeline()` without first calling `update_bvh()`
- `query_pipeline()` will panic if there are pending changes in `changed_colliders` or `removed_colliders`
- All other query methods properly call `update_bvh()` first

### Minor Optimization

**Redundant `changed_colliders` tracking** (line 327-339 in `update_collider_position`)
- When scale changes significantly, the collider is added to `changed_colliders` twice (lines 328 and 339)
- This is harmless but could be optimized by restructuring the conditions

### Positive Changes

✓ Removed complex trait-based conversion system in favor of direct conversions
✓ Proper tracking of collider changes via `changed_colliders`/`removed_colliders` sets
✓ Correct migration from callback-based to iterator-based intersection APIs
✓ Proper addition of `InteractionTestMode::And` to all `InteractionGroups`
✓ Consistent removal of `last_update_frame` parameter throughout

### Recommendation

Add `self.update_bvh();` at the start of the `get_ground()` method (before the `query_pipeline` call) to fix the critical bug.

regenesis-claw (commented):
## Additional Bug Found

You're right, I missed the second bug in `closest_point()`:

**Bug in `closest_point()` method** (line 810):
- The second parameter to `project_point()` is set to `0.0` but should be `f64::MAX`
- This limits the search distance to 0.0, meaning it will never find anything beyond immediate proximity
- The previous commit shows this was deliberately changed from `f64::MAX` to `0.0`

So there are **two bugs** that need fixing:
1. `get_ground()` - missing `self.update_bvh()` call (will panic)
2. `closest_point()` - wrong parameter `0.0` should be `f64::MAX` (won't find anything)

PR #642, opened 2026-03-04
Parcel grass

hukasu:
image


hukasu:
there is no logic for placing the grass yet, so i have a system like this
```rust
fn test(mut commands: Commands) {
    for i in -16..16 {
        for j in -16..16 {
            commands.spawn(ParcelGrass {
                parcel: IVec2::new(i, j),
            });
        }
    }
}
```

PR #693, opened 2026-03-26
feat: late-added UI siblings render on top (matching Unity)

robtfm:
## Summary

- Stamps scene UI entities with a `UiCreationOrder` component (using Bevy's `FrameCount`) when their `UiTransform` is first processed
- Sorts siblings by `(creation_frame, scene_id)` so late-added entities are processed last and their `ChildOf` is inserted last (rendering on top)
- Skips `rightOf` blocking when the target was created in a later frame, preventing late-added entities from reordering existing siblings

Pending clarification of whether the reference client will fix this or maintain it as intended behaviour.


robtfm:
unity matched our behaviour in https://github.com/decentraland/unity-explorer/pull/8178, closing this pr

PR #707, opened 2026-04-08
Nametag bone

hukasu:


robtfm:
i think with the current run animation it's a bit unpleasant. maybe we could keep some 1-second-max transform offset to avoid this?


new
https://github.com/user-attachments/assets/5b4ae0a1-25f3-4ecf-9e05-0be7ed6bd9f4

old
https://github.com/user-attachments/assets/d510e4c4-d317-414c-ab66-35e9a8d72e6e

unity
https://github.com/user-attachments/assets/5dd76d9c-c54b-4190-9e04-3c1852ac7c19

robtfm:
also it has a 1 frame lag which is very obvious at 10fps

https://github.com/user-attachments/assets/f31be87c-a88b-448a-b762-ebf56990c459

to avoid that, you could move the system into `PostUpdateSets`, probably in between `PlayerUpdate` and `AttachSync`

PR #710, opened 2026-04-13
Visibility component update

hukasu:
`VisibilityComponent` now has a `propagate_to_children` field that should be treated

### DIFF
--- a/crates/scene_runner/src/update_world/visibility.rs
+++ b/crates/scene_runner/src/update_world/visibility.rs
@@ -77,6 +77,8 @@ fn visibility_component_on_insert(
     }
 }
 
+/// Removes [`Propagate`] and [`PropagateOver`] when [`VisibilityComponent`] is
+/// replaced or removed.
 fn visibility_component_on_replace(
     trigger: Trigger<OnReplace, VisibilityComponent>,
     mut commands: Commands,
--- a/crates/scene_runner/src/update_world/visibility.rs
+++ b/crates/scene_runner/src/update_world/visibility.rs
@@ -1,5 +1,5 @@
 use bevy::{
-    app::{HierarchyPropagatePlugin, Propagate},
+    app::{HierarchyPropagatePlugin, Propagate, PropagateOver},
     prelude::*,
 };
 #[cfg(not(test))]
@@ -40,16 +40,20 @@ impl From<PbVisibilityComponent> for VisibilityComponent {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Eq, Component)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Component)]
+#[component(immutable)]
+#[require(Visibility)]
 struct AncestorVisibility(Visibility);
 
 fn visibility_component_on_insert(
     trigger: Trigger<OnInsert, VisibilityComponent>,
     mut commands: Commands,
-    mut visibility_components: Query<(&VisibilityComponent, &mut Visibility)>,
+    mut visibility_components: Query<(&VisibilityComponent, &mut Visibility, Option<&ChildOf>)>,
 ) {
     let entity = trigger.target();
-    let Ok((visibility_component, mut visibility)) = visibility_components.get_mut(entity) else {
+    let Ok((visibility_component, mut visibility, maybe_child_of)) =
+        visibility_components.get_mut(entity)
+    else {
         unreachable!("Infallible query.");
     };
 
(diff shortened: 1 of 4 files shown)


hukasu:
result matches with https://github.com/decentraland/sdk7-test-scenes/pull/39

image

crates/scene_runner/src/update_world/visibility.rs
```diff
@@ -1,23 +1,37 @@
...
+};
+#[cfg(not(test))]
 use dcl::interface::ComponentPosition;
-use dcl_component::{proto_components::sdk::components::PbVisibilityComponent, SceneComponentId};
+use dcl_component::proto_components::sdk::components::PbVisibilityComponent;
+#[cfg(not(test))]
+use dcl_component::SceneComponentId;
 
+#[cfg(not(test))]
 use super::AddCrdtInterfaceExt;
```
hukasu: Remove uses of `#[cfg(not(test))]`

crates/scene_runner/src/update_world/visibility.rs
```diff
@@ -26,21 +40,456 @@ impl From<PbVisibilityComponent> for VisibilityComponent {
...
+            .try_insert(Propagate(AncestorVisibility(*visibility)));
+    } else {
+        commands
+            .entity(entity)
+            .try_insert(PropagateOver::<AncestorVisibility>::default());
+    }
+
+    if let Some(child_of) = maybe_child_of {
+        commands.entity(entity).try_insert(child_of.clone());
+    }
```
hukasu: Verify need for reinserting `ChildOf`

crates/scene_runner/src/update_world/visibility.rs
```diff
@@ -26,21 +40,456 @@ impl From<PbVisibilityComponent> for VisibilityComponent {
...
+    };
+
+    *visibility = Visibility::Inherited;
+
+    if visibility_component.propagate_to_children() {
+        commands.entity(entity).try_remove::<(
+            Propagate<AncestorVisibility>,
+            PropagateOver<AncestorVisibility>,
+        )>();
+    }
```
hukasu: Verify correctiness

crates/scene_runner/src/update_world/visibility.rs
```diff
@@ -26,21 +40,456 @@ impl From<PbVisibilityComponent> for VisibilityComponent {
...
+
+    if let Some(child_of) = maybe_child_of {
+        commands.entity(entity).try_insert(child_of.clone());
+    }
```
hukasu, line 95: Verify need of reinserting `ChildOf`

crates/scene_runner/src/update_world/visibility.rs
```diff
@@ -26,21 +40,456 @@ impl From<PbVisibilityComponent> for VisibilityComponent {
...
+            .entity(entity)
+            .try_insert(PropagateOver::<AncestorVisibility>::default());
+    }
+
+    if let Some(child_of) = maybe_child_of {
+        commands.entity(entity).try_insert(child_of.clone());
+    }
+}
+
+fn visibility_component_on_replace(
```
hukasu, line 76: Add comment explaining that this also runs on `OnRemove` and on `OnDespawn`

crates/scene_runner/src/update_world/visibility.rs
hukasu: Needed for the case where `VisibilityComponent` is removed

crates/scene_runner/src/update_world/visibility.rs
```diff
@@ -26,21 +40,555 @@ impl From<PbVisibilityComponent> for VisibilityComponent {
...
-    mut vis: Query<(&VisibilityComponent, &mut Visibility), Changed<VisibilityComponent>>,
-    mut removed: RemovedComponents<VisibilityComponent>,
+    mut visibility_components: Query<(&mut Visibility, Option<&ChildOf>)>,
+) {
+    let entity = trigger.target();
+    let Ok((mut visibility, maybe_child_of)) = visibility_components.get_mut(entity) else {
+        unreachable!("Infallible query.");
+    };
+
+    *visibility = Visibility::Inherited;
```
robtfm, line 86: i think this doesn't work, we need to check AncestorVisibility. consider e.g.

```
- 0: VisiblityComponent(visible: true, propagate_to_children: true)
  - 1: VisiblityComponent(visible: false, propagate_to_children: false)
    - 2:  VisiblityComponent(visible: true, propagate_to_children: true)
```

then `VisibilityComponent.remove(2)`. 2 would take `bevy::Visibility::Inherited` which would be false, when it should be true.

hukasu: if `2` remain in a hierarchy that contains `Propagate`, then `AncestorVisibility` is re-inserted, entering the observer, see new test

PR #720, opened 2026-04-17
Don't limit render distance in preview mode

### LINKED ISSUE #716: Fix preview mode scene visibility when leaving parcels
## Description

In preview mode, the scene disappears when the player leaves the scene parcels. This makes it difficult to test and review scenes while moving around.

## Current Behavior

- Scene distance is currently forced to zero to make stats work properly
- This causes the scene to unload/disappear when the player moves outside the parcel boundaries in preview mode

## Expected Behavior

In preview mode, the scene should remain visible and loaded even when the player moves outside the parcel boundaries, allowing for better testing and review workflows.

## Context

The current behavior is a workaround/hack to get stats working properly. This should be properly fixed to support both correct stats and preview mode scene persistence.

## Related Issues

- #641 — Reduce unnecessary API requests in preview mode

---
Requested by rob via Slack

hukasu:
Closes #716


robtfm:
doesn't seem to work for me? i expect genesis plaza to stay loaded when i leave, and nothing else to load.

https://github.com/user-attachments/assets/8946f55b-e3b4-4384-b9bb-2f1d4ac324d1

hukasu:
if you go to an empty parcel it keeps that scene loaded, if you enter a populated parcel it changes the preview to that parcel

robtfm:
hm ... i think it should just stick with the initial preview scene

hukasu:
how would you change the scene being previewed?

robtfm:
by restarting with the new location.

preview mode isn't really meant for walking around the world, it's meant for inspecting a scene while working on it.

hukasu:
would it make sense disabling `/goto` and `/world` if in preview?

robtfm:
yes, but i don't think it's necessary

robtfm:
is the grass behaviour expected?
Screenshot 2026-04-27 at 16 37 22

robtfm:
i note it works well if i use a cmd line including the position, but if i don't then i end up stuck on the loading screen. does it help if we leave the preview parcel `None` if not specified, and overwrite it once we know the target parcel?

hukasu:
the salmon plane on your world is actually the green plane that extends all the way, but in preview for some reason it does not write alpha(??)

robtfm:
ah yes, it's using the out-of-bounds effect. forgot about that, please disregard.

hukasu:
Now the preview works like it used to, where changing parcels loads the new scene, but you can lock the preview to that parcel with `/lock_preview`

Issue #714, opened 2026-04-14
Store displays in plaza: wearables freeze and show mannequin after animation


regenesis-claw:
### Platform
Not specified (video appears to be desktop/Linux)

### App Version
Not specified

### Device Information
Not specified

### Issue Description
Store displays in the plaza are not rendering wearables correctly. When a wearable loads on a display:
1. The mannequin is hidden and the wearable starts animating
2. A few seconds later, the animation freezes
3. The mannequin reappears

This appears to be an issue with the handling of `show_only_wearables` in `AvatarShape`.

### Expected Behavior
Store displays should continuously show the animated wearable without freezing or showing the underlying mannequin.

### Screenshots / Media
Video demonstration provided in Slack thread (Screencast_20260414_152339.webm)

### Steps to Reproduce
1. Visit the Genesis Plaza
2. Navigate to the store area
3. Observe the wearable displays
4. Wait a few seconds after the wearable loads

### Occurrence
Appears to be consistent based on the report

---

Requested by rob via Slack

hukasu:
tested on Linux, local wasm deployment, decentraland.org and decentraland.zone

PR #727, opened 2026-04-21
Fix body being visible when it should be hidden

hukasu:
image

image


hukasu:
the black mannequins is a different issue tracked in #714

PR #763, opened 2026-05-06
Remove old binaries

hukasu:
Running both the desktop and wasm versions seem to have no issues without them


robtfm:
these are used, but not for the client.

PR #767, opened 2026-05-07
New entrypoint

### LINKED ISSUE #766: Consolidate `main.rs` and `actual_web.rs`
Both files share a lot of code, merge them into one intermediate file

hukasu:
Closes #766 

There was a lot of duplicated code on `main.rs` and `actual_web.rs`, refactor them to decrease duplicated code


src/lib.rs
```diff
@@ -0,0 +1,646 @@
...
+    //         .into_iter()
+    //         .enumerate()
+    //         .map(|(ix, d)| {
+    //             let edge_distance = (1 << ix) as f32 * PARCEL_SIZE;
+    //             let diagonal_distance = (edge_distance * edge_distance * 2.0).sqrt();
+    //             // println!("[{ix}] -> {}", d.max(diagonal_distance));
+    //             d.max(diagonal_distance)
+    //         })
+    //         .collect(),
+}
```
robtfm: i didn't find this logic anywhere else - was it intentionally removed?

src/lib.rs
```diff
@@ -0,0 +1,646 @@
...
+    log::LogPlugin,
+    prelude::*,
+    render::view::RenderLayers,
+};
+#[cfg(target_arch = "wasm32")]
+use bevy::{
+    asset::WasmLoaderHandle,
+    render::{render_resource::PipelineCompilationMode, renderer::RenderDevice, RenderPlugin},
+};
+use bevy_console::AddConsoleCommand;
```
robtfm: we use `console::DoAddConsoleCommand` everywhere else - it was for avoiding console commands that had requirements not included in the test builds. can we switch back to that here please

robtfm (commented):
after starting wasm, my url changed to `http://localhost:3000/?manualParams=true&position=105%2C150&systemScene=basiccontroller.dcl.eth&portables=`, and reloading failed. looks like a bug in the separation of system scene and portables.

robtfm:
one remaining issue: in web previously setting a blank system scene would load the builtin ui (useful when the system scene was broken or unreachable). now it tries to use Some("") as the system scene if you blank it out.

other than that all looks good, you can merge now if you prefer to do that as follow up

hukasu:
done

PR #777, opened 2026-05-12
Skybox reload

### LINKED ISSUE #776: scene.json time setting only takes effect every other reload
## Description

The time-of-day setting in scene.json appears to apply inconsistently during scene reloads. The setting only takes effect every other reload, alternating between applying the configured time and reverting to a default.

## Steps to Reproduce

1. Configure a scene.json with a specific time setting
2. Load the scene in the explorer
3. Reload the scene
4. Observe whether the configured time is applied
5. Reload again
6. Notice the time setting alternates between working and not working

## Expected Behavior

The time-of-day setting in scene.json should consistently apply on every scene reload.

## Actual Behavior

The configured time only takes effect every other reload.

---
Requested by rob via Slack

hukasu:
Closes #776

### DIFF
--- a/crates/visuals/src/day_night.rs
+++ b/crates/visuals/src/day_night.rs
@@ -221,6 +221,7 @@ fn push_time_of_day_from_time_skip(
     ) % TWENTY_FOUR_HOURS;
 
     if time_skip.progress >= 1. {
+        debug!("TimeSkip has ended.");
         commands.remove_resource::<TimeSkip>();
     }
 }
@@ -237,6 +238,7 @@ fn push_time_of_day_from_running_clock(
         } else {
             time_keeper.time + TWENTY_FOUR_HOURS
         };
+        debug!("Starting a TimeSkip from {} to {}.", time_of_day.time, end);
         commands.insert_resource(TimeSkip {
             start: time_of_day.time,
             end,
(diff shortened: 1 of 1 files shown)


crates/visuals/src/day_night/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    let current = time_of_day.time;
+    let delta = time.delta_secs();
+
+    let (mut start, end) = if let Some(skybox_time) = maybe_skybox_time {
+        let target = skybox_time.fixed_time as f32;
+        match (current > target, skybox_time.transition_mode()) {
+            (true, TransitionMode::TmForward) => (current - TWENTY_FOUR_HOURS, target),
+            (true, TransitionMode::TmBackward) => (current, target),
+            (false, TransitionMode::TmForward) => (current, target),
+            (false, TransitionMode::TmBackward) => (current - TWENTY_FOUR_HOURS, target),
```
robtfm:
```suggestion
            (false, TransitionMode::TmBackward) => (current + TWENTY_FOUR_HOURS, target),
```

crates/visuals/src/day_night/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+        (start, target)
+    } else {
+        let target = running_clock.time;
+        let start = if current > target {
+            current - TWENTY_FOUR_HOURS
+        } else {
+            current
+        };
+        (start, target)
+    };
```
robtfm: here i think we should choose the closest direction

robtfm:
also an issue with switching directions. in my scene i set time based on some game state, so i am moving forwards and backwards quite frequently. sometimes it gets the direction wrong and goes round the clock. e.g:

- scene sets 12pm, time reaches 12pm
- scene sets 11am (back), time starts moving towards 11am, time reaches 11:02am
- scene sets 11:01am (forward), time sees 11:01 < 11:02 with dir forwards, moves +23:59

so basically i think we should track target time separately from actual time, and use that as the basis for the transition direction.

PR #790, opened 2026-05-19
Video player audio stream conflict

hukasu:


hukasu:
~~there is a consistent panic right now~~

crates/av/src/av_player_debug.rs
```diff
@@ -146,8 +136,7 @@ type AnyColumn = Or<(
 #[cfg(all(feature = "ffmpeg", feature = "livekit", target_arch = "wasm32"))]
 type AnyColumn = Or<(
     With<AvPlayerColumn>,
-    With<AudioSinkColumn>,
-    With<VideoSinkColumn>,
+    With<VideoPlayerSinks>,
```
robtfm: we never use this feature set, but it uses `VideoPlayerSinks` which doesn't exist. not blocking, maybe worth cleaning up?

crates/av/src/video_player.rs
```diff
@@ -66,134 +74,133 @@ fn init_ffmpeg() {
...
             if let Some(video_sink) = maybe_video_sink {
                 video_sink
                     .command_sender
-                    .send(AVCommand::Repeat(av_player.source.r#loop.unwrap_or(false)))
+                    .send(AVCommand::Repeat(av_player.r#loop()))
                     .report();
                 video_sink.command_sender.send(AVCommand::Pause).report();
             }
-            if let Some(audio_sink) = maybe_audio_sink {
+            if let Some(audio_sink) = mayve_audio_sink {
```
robtfm: added a typo

robtfm (approved):
lgtm pending the typo.

for the record: the pr modifies `max_video_players` to only apply to videos (not to audio streams). the intention is to limit the gpu- and bandwidth-heavy video decoding activity, not to throttle network activity.

PR #801, opened 2026-05-26
Active speaker

hukasu:
Create a PlayerMessage to inform when it is an active speaker

### DIFF
--- a/crates/comms/src/livekit/participant/plugin.rs
+++ b/crates/comms/src/livekit/participant/plugin.rs
@@ -3,7 +3,10 @@ use bevy::{
     platform::collections::HashSet,
     prelude::*,
 };
-use common::{debug_panic, util::AsH160};
+use common::{
+    debug_panic,
+    util::{AsH160, ReportErr},
+};
 use dcl_component::proto_components::kernel::comms::rfc4;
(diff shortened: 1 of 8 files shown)


crates/comms/src/livekit/participant/plugin.rs
```diff
@@ -452,3 +464,113 @@ fn change_volume_of_tracks(
...
+        );
+    };
+
+    let channel = match scene_rooms.get(room.get()).ok() {
+        Some(room) => room.0.clone(),
+        None => "Nearby".to_string(),
+    };
+    for sender in senders.iter() {
+        let _ = sender.send(VoiceMessage {
+            sender_address: format!("{:#x}", participant.identity().as_str().as_h160().unwrap()),
```
robtfm: let's avoid unwrap here, in case auth server or some future non-h160 participant (streamer?) uses a voice channel

crates/comms/src/livekit/participant/plugin.rs
```diff
@@ -452,3 +464,113 @@ fn change_volume_of_tracks(
...
```
robtfm: here also

PR #805, opened 2026-06-02
Alloy

hukasu:
Replace the now 2-year deprecated dependency `ethers` for `alloy`

# TODO
- [x] Fix time not being available on wasm
- [ ] wait for alloy to release a new version with https://github.com/alloy-rs/alloy/pull/4027


robtfm:
Two dependency notes after going through the lockfile diff. The actual build graph only grows by ~15 crates natively and ~11 on wasm; most of the +1k lock lines are cargo locking alloy's unused optional deps (c-kzg/blst, secp256k1, ark-ff via ruint, borsh), which we can't avoid.

**1. `alloy-provider` → `alloy-rpc-client`**

The only consumer is `ethereum_controller.rs`, which opens a ws and calls `raw_request` as a JSON-RPC pass-through, so the provider layer isn't needed. `alloy-rpc-client` with the same `ws-ring` feature covers it:

```toml
# Cargo.toml
alloy-rpc-client = { version = "2.4.2", default-features = false }
# crates/dcl/Cargo.toml
alloy-rpc-client = { workspace = true, features = ["ws-ring"] }
```
```rust
use alloy_rpc_client::{ClientBuilder, RpcClient, WsConnect};

let client: RpcClient = ClientBuilder::default().ws(WsConnect::new(provider_url())).await?;
let result = client.request(Cow::Owned(method.to_owned()), params).await;
```

That drops alloy-provider, alloy-chains, phf 0.14, dashmap, lru, num_enum and async-stream, and puts the wasm dependency graph back to exactly what main has.

**2. reqwest 0.12 → 0.13**

Could you revert this bump? Nothing in the alloy graph needs it (alloy-transport-http isn't pulled in), and it drags hyper-util/typenum/wasm-streams changes into this diff. If there's a reason for it, happy to take it as a separate PR.

PR #813, opened 2026-06-05
Pointer event highlight

### LINKED ISSUE #760: Implement entity highlight on hover (PointerEvents.Info.show_highlight)
## Summary

Protocol PR [decentraland/protocol#224](https://github.com/decentraland/protocol/pull/224) split hover *highlight* off from hover *text*, adding `Info.show_highlight` (default true) alongside the existing `Info.show_feedback`. The proto field is now in tree (PR #758), but bevy-explorer has no entity-highlight rendering at all — so neither flag has any effect on visual outline/glow.

## Current behavior

`crates/scene_runner/src/update_world/pointer_events.rs:hover_text` reads `info.show_feedback` to gate the tooltip text, and ignores `show_highlight` entirely. There is no system that draws an outline or emissive glow on the hovered entity.

## What's needed

1. A renderer hook that draws a hover highlight on the currently-targeted `PointerTargetInfo.container` when at least one in-range pointer entry has `show_highlight != Some(false)`.
2. Gate the highlight on the same distance check used for the tooltip (`passes_distance_check`).
3. Match unity's interpretation of the two flags (verify): `show_feedback=false` should suppress both text and highlight; `show_highlight=false` should suppress only highlight.

## References

- Protocol PR: https://github.com/decentraland/protocol/pull/224
- Unity reference: `ProcessPointerEventsSystem.cs:282-284` reads both flags; `HighlightNewEntity` is the rendering entry point.
- Proto sync into bevy-explorer: #758

hukasu:
Closes #760 

image

image

image

- [ ] Chairs and plants on genesis plaza do not highligh, should they?

### DIFF
--- a/crates/scene_material/src/lib.rs
+++ b/crates/scene_material/src/lib.rs
@@ -2,7 +2,7 @@ use bevy::{
     pbr::{ExtendedMaterial, MaterialExtension},
     platform::collections::{hash_map::Entry, HashMap},
     prelude::*,
-    render::render_resource::{AsBindGroup, Face, ShaderRef},
+    render::render_resource::{AsBindGroup, Face, ShaderDefVal, ShaderRef},
 };
 use boimp::bake::{ImposterBakeMaterialExtension, ImposterBakeMaterialPlugin};
 use common::{structs::PreviewMode, util::InvertedScaleExt};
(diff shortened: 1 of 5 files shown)


crates/scene_material/src/lib.rs
```diff
@@ -277,6 +278,10 @@ impl MaterialExtension for SceneBound {
         }
 
         if let Some(fragment) = descriptor.fragment.as_mut() {
+            fragment.shader_defs.push(ShaderDefVal::UInt(
+                "OUTLINE_GREEN_MESH_TAG".to_owned(),
+                SCENE_MATERIAL_OUTLINE_GREEN_MESH_TAG,
+            ));
```
robtfm, line 284: think this is vestigial?

robtfm:
the meshtag is useful. i think we should also move the cone dither / disable dither, red outline, and inverted scale to use it. that would avoid a lot of messing around with copying materials just to set an extra shader def (at the cost of some shader runtime, but i think it's worth it).

crates/scene_material/src/lib.rs
hukasu: This is to push the value to the shader, so that we don't need to keep in sync the const in the rs and wgsl

PR #821, opened 2026-06-09
Ipfs debug

hukasu:


crates/ipfs/src/lib.rs
```diff
@@ -1194,22 +1238,89 @@ impl AssetReader for IpfsIo {
                 }
             };
 
+            #[cfg(all(feature = "ipfs_debug", not(target_arch = "wasm32")))]
+            let start = std::time::Instant::now();
+            #[cfg(all(feature = "ipfs_debug", target_arch = "wasm32"))]
+            let start = web_time::Instant::now();
```
robtfm: can use web_time on native too (we do this in e.g. scene_runner/lib).

crates/ipfs/src/lib.rs
```diff
@@ -1194,22 +1238,89 @@ impl AssetReader for IpfsIo {
...
+            let maybe_ipfs_path = IpfsPath::new_from_path(path)
+                .map_err(wrap_err)
+                .test_failure(
+                    #[cfg(feature = "ipfs_debug")]
+                    &self.debug_overlay_sender,
+                    #[cfg(feature = "ipfs_debug")]
+                    path,
+                    #[cfg(feature = "ipfs_debug")]
+                    start,
+                )?;
```
robtfm: this block is pretty ugly and repeated many times, maybe worth a macro?

robtfm (commented):
lgtm, just a couple of style points

PR #824, opened 2026-06-10
Scene material key

### LINKED ISSUE #814: use meshtag for instance-specific shader features: cone dither / disable dither, red outline, and inverted scale
the meshtag is useful. i think we should also move the cone dither / disable dither, red outline, and inverted scale to use it. that would avoid a lot of messing around with copying materials just to set an extra shader def (at the cost of some shader runtime, but i think it's worth it).

_Originally posted by @robtfm in https://github.com/decentraland/bevy-explorer/issues/813#issuecomment-4639111391_

hukasu:
Closes #814


robtfm:
remaining issues:

- The baker now calls `mesh_functions::get_tag(in.instance_index)` but does not import `mesh_functions` —`bound_material.wgsl` and `bound_prepass.wgsl` both add `mesh_functions`, to their import block; the baker doesn't. This is a `naga_oil` compose error that surfaces when the imposter-baker pipeline specializes (non-preview mode → imposter baking). It isn't caught by cargo check. And it's gratuitous: mesh_tag is never used in the baker, and the `sample_index = 0u` fallback (line 65) and `@builtin(sample_index)` param (line 48) are also now dead. Simplest fix is to delete the unused mesh_tag/sample_index lines rather than add the import.
- the `playerClicked` RPC send at colliders.rs:191 (used mainly by sdk6 scenes but also available to sdk7 ones) is the sole source feeding RpcCall::SubscribePlayerClicked → the JS playerClicked event (dcl/src/js/events.rs:94), i.e. a scene's onPlayerClick. It fires on input_manager.just_down(IaPointer), but just_down is now only checked on the frame PointerTarget changes

hukasu:
more context on that second point please

robtfm:
RpcCall::SubscribePlayerClicked is the backing for https://docs.decentraland.org/creator/scenes-sdk7/interactivity/deprecated-functions#player-clicks-on-another-player

the `if input_manager.just_down(CommonInputAction::IaPointer, InputPriority::Scene) {
` block which fires the event is now within a block that only runs `if previous_target.as_ref() != pointer_target.0.as_ref().map(|target| &target.container)`

hukasu:
is `if input_manager.just_down(CommonInputAction::IaPointer, InputPriority::Scene) {` the only issue? or is `if native_ui.profile && input_manager.just_down(SystemAction::ShowProfile, InputPriority::AvatarCollider)` also required to run every frame?

robtfm:
the native ui stuff here i don't really care about either way. native ui is still useful sometimes for quick runs but we definitely don't need the extended stuff like viewing avatar profiles on clicks.

PR #835, opened 2026-06-11
Retain highlight on avatar reload

### LINKED ISSUE #828: MeshTags (incl. highlight outline) are lost when an entity's meshes are recreated
## Problem

The selection/hover outline is a `MeshTag` bit (`SCENE_MATERIAL_OUTLINE_GREEN_MESH_TAG`) that `entity_highlighting` writes onto an entity's descendant meshes. When those meshes are **recreated** — a `GltfContainer` reload, an avatar swapping wearables, etc. — the new mesh entities don't carry the tag, so the outline (and any other `MeshTag` state) is lost.

`entity_highlighting` diffs at the **root-entity** level: it only walks descendants for roots that are *newly* in the highlight pass. A reload keeps the same root but replaces its child meshes, so the new meshes are never re-tagged.

## Scope

This is not editor-specific. It affects:
- scene-routed `PointerEvents` `showHighlight` on hover and proximity (a hovered/near gltf that reloads loses its outline),
- **hovered avatars that change wearables** (the avatar's meshes are rebuilt → outline lost),
- the editor's `/highlight` (hit constantly, since it toggles collision masks live),
- any future `MeshTag`-based feature.

The hover case *currently* tends to self-heal only by accident: the reload's empty-frame gap clears `hover_target`, the root expires, and re-acquiring re-tags. Note that #827 (keep the prior gltf instance until the new is render-ready) removes that gap — fixing the flicker but making this highlight loss **deterministic** for hover rather than self-healing. So #827 does not cover this.

## Possible fixes

1. **Narrow:** diff `entity_highlighting` at the mesh level — recompute each highlighted root's descendant mesh set every frame and diff *those* — so reloaded meshes are re-tagged next frame for editor + pointer highlights alike. (This is also the basis for the deferred editor-highlight prune work.)
2. **General (preferred):** persist `MeshTag`s across mesh recreation — when an entity's meshes are respawned (gltf reload, avatar update, …), carry over the existing tag bits to the new meshes. This covers all current and future `MeshTag` uses rather than just the outline.

Surfaced by the component-inspector editor scene; currently worked around scene-side with a clear-then-resend of `/highlight`, which can be removed once this is fixed.

hukasu:
Closes #828

### DIFF
--- a/crates/scene_runner/src/update_world/pointer_events.rs
+++ b/crates/scene_runner/src/update_world/pointer_events.rs
@@ -1,4 +1,5 @@
 use bevy::{
+    app::{HierarchyPropagatePlugin, Propagate},
     ecs::entity::EntityHashSet,
     math::FloatOrd,
     platform::collections::{HashMap, HashSet},
@@ -19,7 +20,7 @@ use dcl_component::{
     },
     SceneComponentId,
 };
-use scene_material::{SceneMaterial, SCENE_MATERIAL_OUTLINE_GREEN_MESH_TAG};
+use scene_material::SCENE_MATERIAL_OUTLINE_GREEN_MESH_TAG;
 
 use crate::{
     renderer_context::RendererSceneContext,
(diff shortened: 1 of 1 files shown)


hukasu:
@robtfm does this fix all of the cases you mentioned on the issue?

robtfm:
will test, looks compelling

hukasu:
i am starting to dislike this `Propagate` solution, i may rollback and do with the previous methodology, but adding the children to the `EntityHashSet`

hukasu:
that last change only works with a change that is on the other branch, which is registering `MeshTag` as a required component of `MeshMaterial3d<SceneMaterial>`, should i make the observers resilient to the `MeshTag` being missing?

robtfm:
You can make a single pr for both changes if that’s easier?

hukasu:
This one is done, merging it without the other one will just momentaniously revert the outline on NPCs

robtfm:
ok so merge this now?

PR #849, opened 2026-06-16
Modify behavior of bounded continuous tweens

### LINKED ISSUE #815: Continuous Tween with duration > 0 diverges from unity-explorer
## Summary

For a **continuous** `PBTween` mode (`RotateContinuous` / `MoveContinuous` / `TextureMoveContinuous`), a non-zero `duration` produces **different behaviour in bevy-explorer than in unity-explorer**. The same component JSON therefore animates differently depending on the client.

The intended model for continuous tweens is speed-driven (the mode's `speed` + `direction` drive the motion). The two clients disagree on what `duration` means in that context.

## bevy-explorer behaviour

`crates/tween/src/lib.rs` — `duration` actively changes a continuous tween:

- Per-frame delta (≈ lines 272–280):
  ```rust
  let delta = if playing {
      if tween.is_continuous() && tween.duration <= 0. {
          time.delta_secs()
      } else {
          time.delta_secs() * 1000.0 / tween.duration
      }
  } else { 0.0 };
  ```
  So a continuous tween with `duration > 0` is **rate-scaled by `1000/duration`** instead of advancing by raw wall-clock seconds.

- Easing (`apply`, ≈ lines 98–104):
  ```rust
  let ease_value = if self.is_continuous() && self.duration <= 0. {
      time
  } else {
      f(time)
  };
  ```
  So `duration > 0` also **re-enables the easing function** for a continuous tween (instead of bypassing it).

In other words, only `duration <= 0` gives "true continuous"; `duration > 0` quietly switches a continuous tween into a rate-scaled, eased mode.

## unity-explorer behaviour

`Explorer/Assets/DCL/SDKComponents/Tween/TweenSDKComponentHelper.cs` — continuous tweens are speed-driven infinite DOTween loops; `duration` is used **only as an optional kill-timeout**:

- Kill once elapsed reaches duration (lines ≈ 145–147, 291–295):
  ```csharp
  if (IsTweenContinuous(pbTween) && TweenSurpassedDuration(pbTween, sdkTweenComponent))
      sdkTweenComponent.CustomTweener.Kill(true);
  ```
  (`TweenSurpassedDuration` fires only when `pbTween.Duration > 0`.)
- Easing is forced to `Linear` for continuous regardless of duration (line ≈ 272):
  ```csharp
  Ease ease = IsTweenContinuous(tweenModel) ? Ease.Linear : GetEase(tweenModel.EasingFunction);
  ```
- `currentTime` is effectively a no-op for continuous.

So in unity, `duration > 0` does **not** change the motion rate or easing — it just gives the continuous tween a finite lifetime (kill after N ms); `duration <= 0` = run forever.

## Impact

A continuous tween authored with a non-zero `duration` (easy to do — e.g. copied from a start→end tween, or a default editor value) will:

- **bevy:** run rate-scaled by `1000/duration` with easing applied, forever.
- **unity:** run at `speed` (linear), then stop after `duration` ms.

Different speed, different easing, and different lifetime for identical content.

## Suggested resolution

Decide the canonical semantics and align bevy with it (likely match unity, since the SDK models continuous as speed-driven):

- Treat `duration` for continuous modes as a kill-timeout only (or ignore it), and
- Always use linear/raw advancement for continuous (don't re-enable easing when `duration > 0`).

Alternatively, if bevy's interpretation is intended, the spec/proto should document it and unity should match — but the two should not silently differ.

Refs: bevy `crates/tween/src/lib.rs` (is_continuous, delta calc, apply/easing); unity `TweenSDKComponentHelper.cs` (kill-timeout, forced Linear ease). Verified against unity HEAD release/2026-06-01.

hukasu:
Closes #815


crates/tween/src/lib.rs
```diff
@@ -91,24 +91,33 @@ impl Tween {
...
+        tween_apply_update: TweenApplyUpdate,
         transform: &mut Transform,
         maybe_mat: Option<&mut PbMaterialComponent>,
     ) {
         let f = self.easing_function();
 
-        let ease_value = if self.is_continuous() && self.duration <= 0. {
-            time
+        let unbounded_continuous = self.is_continuous() && self.duration <= 0.;
+        let ease_value = if unbounded_continuous {
```
robtfm:
```suggestion
        let ease_value = if self.is_continuous() {
```

we should skip easing for all continuous modes (this would behave differently even for default/linear easing, dropping half the delta over the window)

hukasu: the easing function is used when duration is specified, it is a factor over the speed

https://76cfe6fe.adr-cvq.pages.dev/adr/ADR-285

robtfm: then some "constant" easing mode is also needed to be able to apply (e.g.) a constant continuous rotation for a fixed duration, right? there's no way to represent "turn at 90 degrees per second for 8 seconds"

robtfm: i think in this case we should treat the ADR as wrong, unity's behaviour seems to make more sense to me. i think easing a continuous tween is less useful than having a constant tween with a timeout.

PR #867, opened 2026-06-18
Avatars without claimed name uses gray color for point at marker

hukasu:
One of the follow-ups to #752

### DIFF
--- a/crates/avatar/src/point_at_marker.rs
+++ b/crates/avatar/src/point_at_marker.rs
@@ -134,7 +134,14 @@ fn sync_markers(
             continue;
         };
 
-        let bg = name_color(address);
+        let bg = if profile
+            .filter(|profile| !profile.content.has_claimed_name)
+            .is_some()
+        {
+            Color::srgb(0.6, 0.6, 0.6)
+        } else {
+            name_color(address)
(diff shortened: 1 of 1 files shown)


crates/avatar/src/point_at_marker.rs
```diff
@@ -134,7 +134,14 @@ fn sync_markers(
             continue;
         };
 
-        let bg = name_color(address);
+        let bg = if profile
+            .filter(|profile| !profile.content.has_claimed_name)
```
robtfm, line 138: it's a bool - need to check the value, no?

hukasu: if it is `false`, the filter will keep the `Option` as `Some`, which is then checked on the `is_some`

robtfm: ah i thought has_claimed_name was Option<bool>.

ok i would still have said `if profile.is_none_or(|p| !p.content.has_claimed_name) {` but fine

hukasu: if it is `None`, the method returns early, so it would be with `is_some_and`

PR #877, opened 2026-06-19
Custom name color

hukasu:


robtfm:
let's also allow the system scene to clear the value, while still not applying any change unless it is specified.

i think `ClearableColor3 { r, g, b, clear: bool }` so that setting {clear: true} is treated as `Some(None)` and `{r:1}` is treated as `Some(Some(Color3))` probably makes most sense? 

and change the SetAvatarData field to `Option<Option<Color3>>`

PR #886, opened 2026-06-22
Point at visibility

### LINKED ISSUE #753: Add point-at marker visibility setting
Follow-up to #752.

Unity's reference client has a per-user setting for the point-at marker visibility — see `PointAtMarkerVisibilitySettings.cs` and `PointAtMarkerVisibilityController.cs` in unity-explorer. Three options: **All**, **Friends only**, **None**. We currently always show every marker.

## Scope

- New setting key + persistence (DCLPrefs equivalent)
- Plumb the social/friends list through the marker plugin so it can filter remote markers based on the source avatar's friend status (the local player's own marker is unaffected)
- Settings UI entry to flip the value
- Default to **All** to match current behaviour and Unity's default

## Notes

- The avatar's address is already available where we spawn the marker (`crates/avatar/src/point_at_marker.rs::sync_markers`) — the filter would happen at spawn-time and on friend-status changes
- Friends data comes from the social plugin (`crates/social/src/lib.rs`)
- Local player's own marker should always show regardless of setting

hukasu:
Closes #753

# TODO
- [x] Test for friends

### DIFF
--- a/crates/avatar/src/point_at_marker.rs
+++ b/crates/avatar/src/point_at_marker.rs
@@ -13,6 +13,7 @@ use comms::{
 };
 use dcl_component::transform_and_parent::DclTranslation;
 use ethers_core::types::Address;
+use social::{FriendshipState, SocialClient};
 
 use crate::AvatarShape;
 
@@ -338,7 +339,10 @@ fn change_point_at_marker_visibility(
     app_config: Res<AppConfig>,
     marker_overlay: Res<MarkerOverlay>,
     children: Query<&Children>,
-    point_at_markers: Query<&mut Visibility, With<PointAtMarker>>,
+    foreign_players: Query<&ForeignPlayer>,
+    primary_user: Single<Entity, With<PrimaryUser>>,
+    point_at_markers: Query<(&PointAtMarker, &mut Visibility)>,
+    social_client: Res<SocialClient>,
 ) {
     let Some(root) = marker_overlay.root else {
         return;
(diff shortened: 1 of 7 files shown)


crates/avatar/src/point_at_marker.rs
```diff
@@ -1,7 +1,10 @@
 use bevy::{platform::collections::HashMap, prelude::*, ui::UiSystem};
 use common::{
     sets::PostUpdateSets,
-    structs::{PointAtSync, PrimaryCamera, PrimaryUser, ZOrder},
+    structs::{
+        AppConfig, PointAtMarkerVisivbility, PointAtMarkerVisivbilityChanged, PointAtSync,
```
robtfm: `PointAtMarkerVisivbility` -> `PointAtMarkerVisibility`

crates/system_bridge/src/settings/mod.rs
```diff
@@ -184,6 +186,7 @@ impl Plugin for SettingBridgePlugin {
         add_int_setting::<VideoThreadsSetting>(app, &mut settings, &mut schedule, &config);
         add_int_setting::<MaxDownloadsSetting>(app, &mut settings, &mut schedule, &config);
         add_enum_setting::<CacheSizeSetting>(app, &mut settings, &mut schedule, &config);
+        add_enum_setting::<PointAtMarkerVisivbility>(app, &mut settings, &mut schedule, &config);
```
robtfm: just for completeness please add a corresponding entry in system_ui/app_settings.rs as well

robtfm (changes requested):

PR #891, opened 2026-06-23
Lock all invocations of cargo

hukasu:


robtfm:
this is purely defensive right? nothing is floating currently

PR #893, opened 2026-06-23
Use ascending-edge smoothstep in nishita sky/star shader

gonpombo8:
## Summary
Rewrite two reversed-edge `smoothstep(edge0 > edge1, …)` calls in the nishita sky/star shader into the equivalent `1.0 - smoothstep(low, high, x)` form (star brightness + skybox store value). Split out of the editor branch since it touches the production sky/star render path.

## What could break
Behaviour-preserving: `smoothstep(b,a,x) == 1 - smoothstep(a,b,x)`, and reversed edges are well-defined in WGSL (only `edge0 == edge1` is undefined, which doesn't apply). The ascending-edge form avoids relying on reversed-edge behaviour across backends. Renders for all players, so worth a visual sanity check of the sky/stars.

## How to test
`cargo check`; render a scene and eyeball the skybox + stars at dawn/dusk.


robtfm:
taken into #894 directly to avoid merge conflicts

PR #895, opened 2026-06-24
Gltf container change clean up

### LINKED ISSUE #827: GltfContainer reload flickers — keep prior instance until the new is render-ready
## Problem

Changing a `GltfContainer` re-instances the whole gltf: the old child meshes are despawned and new ones spawned, leaving a visible gap/flicker for the frame(s) until the new instance is render-ready. This happens even for changes that don't affect geometry at all — e.g. a collision-mask-only change (`visibleMeshesCollisionMask` / `invisibleMeshesCollisionMask`).

## Where it surfaced

The component-inspector editor scene toggles collision masks **live** while editing (it overlays an editor pick-collider layer onto everything, and strips `CL_PHYSICS` from the selected entity while a gizmo is active so dragging it can't shove the player). Each toggle triggers a full gltf reload, so the flicker is frequent and obvious.

It also drops the editor's selection outline: `entity_highlighting` diffs at the **root-entity** level, so the reloaded child meshes don't get re-tagged. That part is currently worked around scene-side, but it goes away once the reload does.

## Proposed approach

Mirror how **materials** are handled — keep the prior gltf instance rendering until the new one is render-ready, then swap atomically — so there's never an empty/half-loaded frame. (A mask-only delta could additionally be applied in place without any reload, but the deferred-swap covers all reload cases.)

hukasu:
Closes #827

### DIFF
--- a/crates/scene_runner/src/update_world/gltf_container.rs
+++ b/crates/scene_runner/src/update_world/gltf_container.rs
@@ -114,6 +114,8 @@ impl Plugin for GltfDefinitionPlugin {
                 .before(TransformSystem::TransformPropagate),
         );
         app.add_systems(Update, debug_modifiers);
+
+        app.add_observer(clear_stale_instance);
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/scene_runner/src/update_world/gltf_container.rs
robtfm: this exit path needs a `remove::<GtlfReady>` now for consistency

crates/scene_runner/src/update_world/gltf_container.rs
```diff
@@ -214,22 +220,9 @@ fn update_gltf(
...
 
-    for (ent, scene_ent, gltf, maybe_loaded, maybe_processed) in new_gltfs.iter() {
+    for (ent, scene_ent, gltf) in new_gltfs.iter() {
         debug!("{} has {}", scene_ent.id, gltf.0.src);
 
-        if let Some(GltfLoaded(Some(instance))) = maybe_loaded {
-            // clean up from loaded state
-            // instances_to_despawn_when_ready.push(*instance);
-            scene_spawner.despawn_instance(*instance);
-        }
```
robtfm, line 224: this instance is no longer despawned here. `GltfLoading(instance)` + `Changed<GltfDefintion>` -> instance stays forever

crates/scene_runner/src/update_world/gltf_container.rs
```diff
@@ -214,27 +229,20 @@ fn update_gltf(
...
-            instances_to_despawn_when_ready.push(*instance);
-        }
         commands
             .entity(ent)
             .remove::<GltfLoaded>()
             .remove::<GltfProcessed>();
 
+        if let Some(GltfLoaded(Some(instance_id))) = maybe_loaded {
+            if !has_gltf_processed {
+                commands.entity(ent).try_insert(GltfReady(*instance_id));
```
robtfm: i think just despawn the loading one immediately like we did before.

this way it will also replace/despawn the previous one.

- gltf A fully loaded -> GltfReady(a)
- gltf B start loading
- gltf C start loading -> GltfReady(b) -> despawn a -> nothing visible
- gltf C finish loading -> GltfReady(c) -> despawn b

PR #900, opened 2026-06-25
Hide nametag

hukasu:


crates/scene_runner/src/update_world/avatar_modifier_area.rs
```diff
@@ -352,6 +366,15 @@ fn update_avatar_modifier_area(
 
         modifiers.areas = areas_clone;
 
+        if has_hidden_nametag ^ needs_hidden_nametag {
```
robtfm: nit: i think != is clearer

crates/scene_runner/src/update_world/avatar_modifier_area.rs
```diff
@@ -255,6 +264,11 @@ fn update_avatar_modifier_area(
                         .modifiers
                         .contains(&(AvatarModifierType::AmtDisablePassports as i32));
                 }
+
+                needs_hidden_nametag |= area
+                    .0
+                    .modifiers
+                    .contains(&(AvatarModifierType::AmtHideNametags as i32));
```
robtfm: let's put this in the permit block, we can also change the permission strings (3x in structs.rs) to reflect that it is for hiding "avatars/nametags"

robtfm:
please update the permission strings too - mentioned in prev comment

PR #901, opened 2026-06-26
Add react-web: React DOM HUD port over a bridge scene

gonpombo8:
## Summary

Replaces the old boot page with the **React DOM HUD as the production web page**. Merging this deploys it: CI builds the React app + bridge scene + wasm into the one `deploy/web` publish (npm + CDN), served at the explorer URL.

**What's in it**
- **React HUD** (`react-web/`): login/loading, destination picker, chat + nametag bubbles, sidebar, map (satellite atlas), backpack, emotes, communities, friends, notifications, settings, gallery, profile — driven over a headless SDK7 **bridge scene** relaying `SystemApi` ↔ `BroadcastChannel`.
- **Production layout** (`deploy/web/`): React page at the root; the old engine boot page moved to `engine/` (loaded in a same-origin iframe); the exported bridge-scene realm at `bridge-scene/static/`; the shared root-scope service worker.
- **CI** (`build-deploy-web`): wasm → `engine/pkg`, prebuild stamps `PUBLIC_URL`, React builds with the versioned-CDN base, bridge scene exports with a CDN-pinned content `baseUrl`, then oddish publishes.
- **URL params** (parity with the old page): `?position=x,y`, `?realm=name.dcl.eth` (both skip the picker; bad realms get a dismissable "World not found"), `?portables`, `?preview`, plus passthrough of scene-level `SceneParams`.

**Serving rules (the hard-won part)**
- The entry URL has **no trailing slash** (`/bevy-web`) while assets live on the **versioned CDN** → `index.html` refs are absolute (`PUBLIC_URL`); runtime same-origin pieces (engine iframe, bridge scene, SW) resolve against the page directory (`PAGE_DIR`), never the CDN.
- The **service worker** rewrites navigations to `COEP: credentialless` (catalyst ``s send no CORP). The page canonicalizes to the directory URL and does a one-shot reload to enter the SW scope — `serviceWorker.ready` can't be used (it matches the out-of-scope creation URL).
- SW + `wasm_fs` share the `ipfs-path-cache-v1` cache (raw fetch in, processed BC7 out). A branch-local v2 bump broke that pairing and was reverted; both sides now carry a comment pinning the name.

## What could break

- **This changes the production page.** The old boot page remains available at `engine/index.html` (npm-package consumers bump their entry path).
- The engine iframe is now **SW-controlled** in production for the first time (the old no-slash entry never was) — the ipfs/processed-asset cache path actually runs. Watched for regressions locally; worth watching on zone after deploy.
- First visit does one automatic reload (SW takeover). Hard reloads bypass the SW by design and self-heal on the next navigation.
- Non-Chromium desktop and mobile get gate pages (same rules as the engine's own gate).
- Known, pre-existing: `marketing-files.decentraland.org` sends a hardcoded `ACAO: decentraland.org` (broken for zone too — needs an infra ticket).

## How to test

```bash
# tier 1 + visual (no engine needed)
cd react-web && npm ci && npm test && npm run test:visual

# prod-shape, full engine (needs a local wasm build)
wasm-pack build --target web --out-dir ./deploy/web/engine/pkg --no-default-features --features="livekit,social"
cd react-web && PUBLIC_URL= npm run build && cd bridge-scene && npm run bundle
npx serve deploy/web   # then open localhost:3000 → guest → world
```

- `?position=-1,-1` / `?realm=boedo.dcl.eth` → straight to loading, no picker; `?realm=bogus.dcl.eth` → "World not found" → OK → picker.
- After merge: verify `decentraland.zone/bevy-web` (real CDN + headers) before rollout.


gonpombo8:
## Production-readiness pass

Pushed the permission dialog + loading fix + a build/perf/code hardening sweep. Summary of analysis and what was changed vs. what's left as follow-ups.

### Build size
Split the single ~880 KB bundle into cacheable, parallel chunks (vite `manualChunks`):

| chunk | raw | gzip | loads |
|---|---|---|---|
| index (core HUD) | 92 KB | 31 KB | always |
| vendor (react) | 143 KB | 46 KB | always, rarely changes |
| menus (full-screen pages) | 89 KB | 28 KB | always |
| emoji (dataset) | 533 KB | 46 KB | isolated, cached |
| design | 20 KB | 7 KB | always |
| Showcase | 11 KB | 4 KB | **dev-only (`?showcase=1`)** |

- **Core HUD chunk 625 KB → 92 KB.** Biggest finding: `chat/emojis_complete.json` is **716 KB (~half the JS)**, pulled into core by `Chat.tsx` autocomplete; isolated into its own cached chunk. Total transfer is unchanged (same code) but now splits + caches well, and the engine WASM dominates first paint regardless.
- Lazy-loaded the dev-only Showcase so it never ships in the prod path.
- `.vite` was untracked and would've been committed — now gitignored.

### Performance
- Per-frame bridge systems (`proximity`, `pointer`, `nametags`, `world` mic, `avatarPreview`) are already change-gated/throttled — no per-frame `postMessage` churn. ✅
- BroadcastChannel transport: listeners are deduped Sets with correct cleanup. ✅
- **Fixed:** `EngineDriver` scheduled a 6 s ready-fallback `setTimeout` never cleared on `dispose()` → timer leak / late emit on logout. Now tracked + cleared.

### Code correctness (fixed)
- `PhotoDetail` clipboard write had no `.catch()` → silent failure. Added.
- `ProfileCard` copy-reset `setTimeout` lived in a callback (could fire post-unmount) → moved into an effect with cleanup.
- Two pre-existing **bridge-scene lint errors** (`session.ts` never-typed template, `nametags.tsx` void-arrow) fixed; bridge-scene lint now clean.
- Corrected a misleading `usePlaces` comment (it uses a `cancelled` flag, not AbortController — deliberate, because `placesApi` shares/caches promises and aborting would break other consumers).

### Dead-code analysis
Ran `knip`. Signal is **noisy here and mostly false positives**: it flags the entire `bridge-scene/` (a separate SDK package with its own entry) and all `protocol.ts` union-member interfaces (used structurally in the `PageToScene`/`SceneToPage` unions) and unused design-system primitives/glyphs (intentional library surface). No safe automated deletions; left as-is.

### Verification
App + bridge-scene typecheck clean · bridge-scene lint clean · **188 tests pass** · build clean (no warnings).

### Recommended follow-ups (not done — flagged)
1. **Defer the emoji dataset** (~46 KB gz off initial load): lazy the picker + extract the `:shortcode:` autocomplete into a `React.lazy` boundary. Skipped now — it's coupled to Chat's Enter-to-apply keyboard handling and is tested; wants its own change + QA.
2. **Communities invitations & requests** — not mocked, **unimplemented**. The "Invites & Requests" sidebar is an empty placeholder and invited communities show JOIN instead of Accept/Reject. API exists (`GET /members/{me}/requests?type=invite|request_to_join`, `PATCH /communities/{id}/requests/{requestId}` `{intention}`).
3. **Places Favourites** tab returns `[]` (needs signed fetch); **community thumbnail upload** on create is deferred (previews locally, not sent).
4. `useEngineSession` returns one large object recreated each render; consumers aren't memoized. Low real-world impact (bridge side is change-gated) but the architectural lever if re-render cost shows up — split into per-domain contexts / memoize slices + `React.memo` consumers.

gonpombo8:
**Update — permission dialog approach changed.** Reverted the React permission relay from my earlier comment in favour of the engine's **native** permission dialog: `web.rs` now sets `NativeUi.permissions = true` for the web build (one flag; other native UI stays off, React still owns the rest of the HUD). The native dialog already renders every `PermissionType` and also handles the "permission used" toasts + Settings permissions tab, which the React port didn't reimplement.

Dropped: the bridge `permissions` domain, the protocol messages, the React dialog + session slice, and the `scene_name`/`realm` fields I had added to the engine's bridge `PermissionRequest` (Rust reverted). The build-size, loading-flash, and correctness fixes from the previous comment are unchanged.

Worth a quick manual check on a real build: that the native dialog **receives clicks** through the React DOM overlay on top of the engine iframe.

PR #912, opened 2026-06-30
Run the react-web HUD on native bevy (macOS)

gonpombo8:
## Summary

Hosts the react-web HUD over the native engine as a transparent webview. The SDK7 bridge-scene feeds it data exactly like on web — on native there's no browser `BroadcastChannel`, so it's polyfilled over a new `SystemApi` bridge (`BridgeToPage` / `GetBridgeStream` + ops). Login → world, chat, scene-loading, and per-pixel input passthrough work on macOS. Behind the off-by-default `react-hud` feature.

## Two approaches (for follow-up)

1. **macOS overlay (this PR).** The webview sits in its own borderless transparent `NSWindow` over bevy's window, created via objc. It works *because the overlay window is outside bevy* — bevy never renders to it, never manages its cursor.

2. **Second bevy window (tried, doesn't work).** Host the webview in a second bevy window using bevy's portable `transparent` / `window_level` / `cursor_options.hit_test`. This would be cross-platform, but it **crashes on macOS**: `bevy_winit::update_cursors` calls winit `set_cursor` on the new window, which compares against a nil `NSCursor` and panics under objc2 (a winit secondary-window bug). We don't even reach compositing — and bevy creates + presents a GPU surface for *every* window, which would likely cover the webview anyway. So cross-platform needs upstream winit/bevy fixes, or per-OS overlay code. Left as follow-up.

## What could break

macOS-only and feature-gated, so default builds are untouched. The new `SystemApi` variants are emitted always but only consumed under `react-hud` (no-op otherwise).

## How to test

```
# serve react-web (vite) + the bridge-scene, then:
cargo run --no-default-features --features react-hud --bin decentra-bevy -- \
  --server <realm> --ui <bridge-scene url>
```
The webview loads `REACT_HUD_URL` (defaults to the vite dev server with `?native=1`).


robtfm:
superceded by #954

PR #915, opened 2026-07-01
fix/react-web-hud-01: nearby avatar → profile-card + radial hover tooltips + stackable popups

pabloes:
Targets `main` (the `feat/react-web-hud` stack below merged via #901/#907). Clicking a nearby avatar in the world now opens their **profile-card** at the cursor. Also reworks the free-cursor hover prompts into the old scene's radial layout, and introduces an imperative popup layer that the card, passport and confirms all run on.

## What's included

- **Nearby avatar → profile-card** (`bridge-scene/domains/avatarPointer.ts`): SDK7 pointer-down on each `PlayerIdentityData` avatar (hover "Show Profile"). Only the **address** is forwarded to React; the card is anchored at the live DOM cursor.
- **Full action parity** in `ProfileCard`: relationship-driven friend CTA (Add / Accept + Reject / Requested / Unblock), **View Passport**, **Mention**, **Block** (confirm) / **Unblock**. Consolidated `onAddFriend`/`onBlock` into a single `onFriendAction` across ProfileCard/Chat/FriendsPanel/App, and split the card into a `ProfileCard` container (resolves identity + relationship from the session) plus a presentational `ProfileCardPresentation`.
- **Imperative popup layer** (`src/design/popups.tsx`): `openPopup` / `showDialog` / `showConfirm` over a module store, rendered by a single `<PopupHost/>` (read via `useSyncExternalStore`). A popup can be opened from anywhere and can open another (card, passport, confirm), no prop-threading. The world card, the passport and the Block confirm all run on it. Escape is engine-driven: the engine's `Cancel` system action, relayed by the bridge, closes the topmost popup, so it works even while the engine holds keyboard focus.
- **Radial hover tooltips**: free-cursor world-hover prompts spread into radial slots around the pointer (right, left, top, up to 7); pointer-locked keeps the stack under the reticle. Cursor position is **DOM-owned** (a `window` mousemove store read via `useSyncExternalStore`), so no coordinates travel over the bridge and only `<Pointer>` re-renders per move. Out-of-range prompts distinguish camera vs player distance (glyph + copy).
- **Cleanups**: the `avatarClick` x/y carrier and the `hoverPos` wire message were removed (the DOM owns the cursor now); the earlier `display:none` modal-stacking hack (and its `r.width === 0` clamp guard) is gone now the confirm renders as its own popup.

## Review follow-ups (gonpombo8) — addressed in-branch

- **Privacy gating**: the avatar pointer-down respects `AvatarModifierArea` `DISABLE_PASSPORTS` (and `HIDE_AVATARS`) via `getAvatarModifiers`, both at rescan and at click time. The poll runs on two dt-throttled, single-flight systems (bevy-ui-scene parity), not a frame counter.
- **Self-filter**: the local player's own avatar never gets the "Show Profile" handler.
- Each avatar click opens a **fresh** card popup instance (no state bleed when switching avatars); the card and passport receive the full `relationship` (no duplicate request on an incoming request; ADD FRIEND hidden for blocked/incoming).
- World card now shows the player's **photo** (resolved by address from the nearby roster / friends / fetched passports via `resolveIdentity`).
- **Block** now shows a confirm. **Report** was dropped entirely (a `console.log` stub with no backend, so a dead action was worse than hiding it). Re-add when a moderation endpoint exists (backlog #22).
- `--text-dim` defined in `tokens.css` (was a phantom fallback).

## Testing

- `npm run typecheck` (react-web + bridge-scene), `npm test` — **226 unit tests** pass (`profileCard.test.tsx` / `profileCardContainer.test.tsx` cover every action; `popups.test.tsx` covers the popup layer; session covers mention).
- `npm run test:visual` — new tier-1.5 baselines `profile-card.png` + `hover-tooltips.png` (seeded via `?mock=1&simhover=N`). Also fixes `enterWorld` in the visual spec, which was failing the whole suite after the destination-picker change.
- **Visual suite made deterministic** (7802233a): the harness's `clock.setFixedTime` pinned `Date` but left `setTimeout` on real time, so screenshots raced the mock bridge's staggered delivery (pass/fail depended on machine load). Switched to `clock.install` + `fastForward(15s)` so every test captures the same fully-delivered state (suite drops ~2min to ~35s). Also blocks the COI service worker in tests (its first-visit reload from #928 was killing the gate tests), and regenerates the 4 baselines that were frozen mid-load. Adds `.nvmrc` (24), the #926 toolchain needs Node >= 20.19.
- Merged latest `main` (#927 dropped the orphaned `version.ts` commit stamp; #953). No `COMMIT_HASH` consumers remain.

## Manual verification (world-space, can't be mocked — see `review.md §4`)

- [ ] Aim at a real avatar, "Show Profile" hover, click frees the cursor and opens the card at the click point.
- [ ] Inside a `DISABLE_PASSPORTS` avatar-modifier area, aiming at an avatar shows **no** "Show Profile" and clicking opens nothing.
- [ ] Radial hover prompts sit around the mouse at a non-100% HUD scale (coordinate space / `--ui-scale`).

## Notes / deferred

- **Invite to Community**: the earlier orphaned plumbing (protocol messages, session `invitable`/`requestInvitable`/`invite`, bridge handlers) was **removed** (no UI consumer). Rebuild fresh against the members `/invites` endpoint when communities is revisited (backlog #17b).
- **Report**: removed (see above); backlog #22 to re-add with the endpoint.
- Two P2 items from the review are parked in the backlog: popup Promise not settling on engine-`Cancel` (#7b), and chat name showing the raw address for players who left nearby range (#26).
- Backlog items 18–20 (Context/store, deep-linkable URLs, voice "who's speaking") are documented, not addressed here.


gonpombo8 (commented):
## Review — nearby avatar → profile-card + radial hover tooltips

Clicking a nearby avatar opens their `ProfileCard` at the cursor with full action parity, plus a radial free-cursor hover layout. Migration hygiene is strong — `tsc --noEmit` clean, all `ProfileCard` prop/message/harness wiring checks out, the `onAddFriend`/`onBlock` → `onFriendAction` consolidation is complete, and tier-1 + tier-1.5 tests/baselines and the `review.md`/backlog updates are wired. Findings that survived verification:

### 🔴 Correctness

1. **`bridge-scene/src/domains/avatarPointer.ts:22` — the "Show Profile" click ignores `AvatarModifierArea` `DISABLE_PASSPORTS`.** It registers a pointer-down on every avatar unconditionally. The engine models the modifier (`avatar_modifier_area.rs:265`, `AmtDisablePassports → hide_profile`) and the old `bevy-ui-scene` respected it (`avatar-tracker`'s `isProfileBlocked`), so **inside a passport-disabled privacy area, clicking an avatar still opens the profile card** — a privacy/parity regression. Needs to gate registration/click on the passport-disable state (per-avatar), the way nametags already consult presence.

2. **`react-web/src/App.tsx:253` — the world `<ProfileCard>` has no `key`, so switching avatars reuses the instance and bleeds state.** Click A → "ADD FRIEND" (arms `justSent` + a ~1100 ms auto-close timer); click B in-world before it fires (an engine event — no scrim dismiss) → `setWorldCard(B)` reconciles the same element → B shows a **false "✓ REQUEST SENT"**, and A's timer then dismisses B. `ProfilePassport` 10 lines up uses `key={address}` for exactly this.

3. **`react-web/src/App.tsx:246` — `ProfilePassport` gets reduced `isFriend`/`requested`, not `relationshipOf`.** For an **incoming** request both are false → the passport shows a live "ADD FRIEND"; clicking fires a **duplicate request to someone who already requested you** (blocked users show it too). Now reachable via the new world card's "View Passport".

4. **`bridge-scene/src/domains/avatarPointer.ts:18` — no self-filter.** The guard is only `data.address === ''`, but the local player's avatar carries `PlayerIdentityData` → you can click your own avatar in third-person → a **degenerate empty card** (`isMe` → no menu/actions). `nametags.tsx` excludes self; this doesn't.

5. **`bridge-scene/src/domains/communities.ts:154` — `getInvitableCommunities` types the response as a bare array while every sibling social-api call in the file unwraps `{ results: [...] }`.** If `/invites` is likewise enveloped, `invitable[address]` stores the wrapper → `invitableCommunities.length` is `undefined` → the "Invite to Community" row never appears (and `.map` would throw). The mock returns a bare array, so tests pass while the real feature is dead. *(Confirm the real endpoint shape — the intra-file inconsistency is the tell.)*

### 🟡 Performance

6. **`bridge-scene/src/domains/pointer.ts:34` / `useEngineSession.ts:376` — `hoverPos` re-renders the whole HUD per mouse-move frame.** While the cursor is free and hovering, each move sends `hoverPos` → `setHoverPos` returns a fresh `session` object → full HUD reconcile, though only `<Pointer>` consumes it (new amplification — hover previously fired only on enter/exit).

7. **`ProfileCard.tsx:181` / `requestInvitable` — no `fetchedRef` cache guard** (unlike getProfile/getNotifications) → re-opening the same card re-issues the signed `/invites` GET each time, though the answer is already cached in `invitable[address]`.

### 🟢 UX / design-system

8. **Block fires immediately with no confirm while Report has a confirm** → asymmetric safety for two adjacent destructive actions (a mis-click Block is unrecoverable). The Report confirm also hides the card via inline `display:none` (a workaround the code itself flags → backlog #9).

9. **Duplication / tokens** — `ProfileCard` inlines a 3rd copy of the relationship friend-CTA (backlog #3 wants a shared `FriendButton`/`useRelationship`; ProfilePassport & CommunityModal keep diverging copies), hand-rolls an action menu `ContextMenu` already models, and references a **phantom `--text-dim`** token (undefined in `tokens.css` → always the raw `rgba` fallback; canonical is `--ink-6/7`).

*Checked & cleared (not flagged):* coordinate anchoring is **not** a DPR bug (engine `screenCoordinates` are CSS px; the card is portaled outside `--ui-scale`), and the recycled-entity click handler re-reads the address fresh (no stale-click).

react-web/src/features/session/useEngineSession.ts
```diff
@@ -359,13 +402,24 @@ export function useEngineSession(createDriver: () => LoginDriver): EngineSession
...
         case 'cursorLock':
           setCursorLocked(msg.locked)
           break
         case 'proximity':
           setProximity(msg.tips)
           break
+        case 'avatarClick':
+          // The card's scrim swallows mouse input, so the engine's raycast freezes and never sends
+          // the hover-exit — clear the hover here or its tooltip stays painted beside the card.
+          setHover([])
```
gonpombo8, line 400: **[blocking] The hover clear here is one-sided, and the other ProfileCard scrims don't clear hover at all.**

The scrim freezes the engine raycast so no hover-exit ever fires. Consequences:

1. Close the avatar card while still aiming at the same avatar → no "Show Profile" tooltip until the cursor leaves and re-enters (the engine never saw an exit, so no new enter edge arrives).
2. The chat and friends-panel cards (`Chat.tsx:592`, `FriendsPanel.tsx:290`) open the identical scrim with **no** `setHover([])`, and `<Pointer>` stays mounted (`App.tsx` only hides it on `pageOpen`), so a frozen world-hover tooltip stays painted next to those cards.

This is a property of every scrim, not of avatar clicks — please fix at the mechanism level (clear/suppress hover whenever any scrim is open) instead of per-message-case.

react-web/bridge-scene/src/domains/avatarPointer.ts
```diff
@@ -0,0 +1,67 @@
...
+  // HIDE_AVATARS — the engine only hides the mesh (the collider stays live), so without this an
+  // invisible avatar would still show "Show Profile" over empty space and leak who's there. A
+  // disabled avatar is treated as absent in the rescan, so its pointer-down (and the hover) is
+  // removed within a poll cycle; the click-time check stays as a belt-and-braces for that stale window.
+  const passportDisabled = new Set<string>()
+  let frame = 0
+  ctx.push(() => {
+    // Avatars enter/leave constantly; rescan ~2×/s (PlayerIdentityData is the engine's source of
+    // truth, same as nametags) and diff against what already has a handler.
+    if (frame++ % 30 !== 0) return
```
gonpombo8: **[blocking] The frame logic — three problems in one block:**

1. `frame++ % 30` is a frame counter pretending to be a clock. Systems receive `dt`; counting frames makes the poll cadence a function of the player's refresh rate — a 144Hz player polls 2.4× as often as a 60Hz one, and a stuttering client polls *slower* exactly when the privacy window matters most. Use a `dt` accumulator.
2. The modifier RPC poll and the entity rescan are two unrelated cadences welded to one counter — split them.
3. The fire-and-forget `void …then()` has no in-flight guard; it's safe only because the engine channel happens to be FIFO — an invariant living three layers away in Rust. Add a `polling` flag so the safety is local.

Also prefer rebuilding the set over `clear()`+refill (atomic by construction):

```ts
let passportDisabled = new Set<string>()
let cooldown = 0
let polling = false
ctx.push((dt) => {
  cooldown -= dt
  if (cooldown > 0 || polling) return
  cooldown = 0.5
  polling = true
  BevyApi.getAvatarModifiers()
    .then((mods) => {
      passportDisabled = new Set(
        mods.filter((m) => m.hideProfile || m.hideAvatar).map((m) => m.userId.toLowerCase())
      )
    })
    .catch(() => {}) // transient; next poll retries
    .finally(() => { polling = false })
})
```

(For contrast, the `pointer.ts` system is exactly right — edge-gated, no I/O.)

react-web/src/features/chat/ProfileCard.tsx
```diff
@@ -141,10 +167,15 @@ export function ProfileCard({
...
+  const hasDestructive = (!!onFriendAction && !!user.address) || !!onReport
+  const hasMenu = !isMe && (!!onViewProfile || !!onMention || hasDestructive)
+
   return createPortal(
     <>
-      <div className={styles.scrim} onClick={onClose} />
-      <div ref={cardRef} className={styles.card} style={{ left: pos.left, top: pos.top }} onClick={(e) => e.stopPropagation()} role="dialog" aria-label="Profile">
+      {/* Hide the card (not unmount — keeps its state) while a destructive confirm is up, so the
+          popup isn't stuck behind it. Proper stacking is backlog item 9 (consolidate modals + z-layer). */}
+      <div className={styles.scrim} onClick={onClose} style={confirmReport || confirmBlock ? { display: 'none' } : undefined} />
```
gonpombo8: **[blocking] The modal-stacking hack is already self-replicating.**

Hiding the card/scrim with inline `display: 'none'` while a confirm is up forced a second hack: the `r.width === 0` guard in the clamp effect above. Every future dialog opened from the card must be threaded into both style ternaries **and** remembered against the rect guard — forget one flag and the confirm renders under the card.

The z-layer tokens exist — render the confirm one layer above instead of hiding the parent (or land backlog item 9's modal consolidation here).

react-web/src/features/chat/ProfileCard.tsx
```diff
@@ -164,42 +195,102 @@ export function ProfileCard({
...
+                <ReportIcon />
+                <span>Report</span>
+              </button>
             )}
           </div>
         )}
       </div>
+      {confirmReport && (
+        <ModalShell
+          title={`Report ${base}?`}
```
gonpombo8: **[blocking] Report and Block confirms are copy-pasted `ModalShell` blocks** differing only in strings and callback, growing the rebuilt-confirm count in the codebase from 2 to 4 (ExitConfirm, WorldVisitModal already roll their own).

Minimal fix: one `confirm: 'report' | 'block' | null` state driving a single parameterized ModalShell — which also collapses the doubled `confirmReport || confirmBlock` ternaries on the scrim/card. (Backlog item 7's `ConfirmDialog`/`useConfirm` primitive is the real home.)

react-web/src/engine/protocol.ts
```diff
@@ -652,11 +680,28 @@ export interface HoverAction {
...
   actions: HoverAction[]
+  /** Cursor screen-pixel position (from PrimaryPointerInfo) so the hint sits at the pointer, not the
+   *  reticle. It's the screen centre while pointer-locked; absent if unavailable. */
+  x?: number
+  y?: number
+}
+
+/** Streamed cursor position while a hover is active (free cursor) so the tooltip follows the mouse —
+ *  the hover stream only fires on enter/exit, so position updates come through here per-frame. */
+export interface HoverPosMessage {
```
gonpombo8: **[blocking] Cursor coordinates now cross the bridge three ways** — `HoverMessage.x/y`, this per-frame `HoverPosMessage`, and `AvatarClickMessage.x/y` — with `pointer.ts` and `avatarPointer.ts` each sampling `PrimaryPointerInfo` independently, and the mock emulating two of them separately. Yet all three land in the same `hoverPosRef` store on the page side.

One cursor-position channel + purely semantic events (hover actions, avatar-clicked) stops the next pointer-anchored feature (context menu, drag, right-click) from adding a fourth carrier.

react-web/src/features/pointer/Pointer.tsx
```diff
@@ -58,16 +130,22 @@ export function Pointer({ hover, locked, proximity }: { hover: HoverAction[]; lo
...
-              <KeyCap button={a.button} />
-              <span className={styles.hintText}>{a.enabled ? a.text : 'Too far, get closer'}</span>
-            </div>
-          ))}
-        </div>
-      )}
+      {active &&
+        (!showReticle && hoverPos ? (
+          // Free cursor: spread the prompts radially around the pointer (up to 7 slots).
+          <div className={styles.hintsRadial} style={{ left: hoverPos.x, top: hoverPos.y }}>
```
gonpombo8: **[non-blocking] Radial hover prompts have no viewport clamping.** The root is `overflow: hidden`, so near the right edge the *primary* (right-middle) slot clips off-screen entirely; left slots (`translate(-100%)`) clip at the left edge, and the top slot within a ~68px band of the top — wider at `--ui-scale > 1`.

This same PR clamps the ProfileCard to the viewport using its real size — same treatment here.

react-web/src/App.tsx
```diff
@@ -241,12 +256,26 @@ function Hud(): React.JSX.Element {
...
-              requested={session.friends.sent.some((r) => r.address.toLowerCase() === passport.address.toLowerCase())}
+              relationship={relationshipOf(passport.address)}
               onAddFriend={(address) => session.friends.act('request', address)}
               onClose={() => setPassport(null)}
             />
           )}
+          {session.worldCard && (
+            <ProfileCard
+              key={session.worldCard.address}
+              user={{ address: session.worldCard.address, name: session.worldCard.name }}
```
gonpombo8: **[non-blocking] The world profile card never shows the player's photo.** `avatarClick` carries only address/name, so this always renders the initials fallback — while the same player opened from chat shows their picture (via `pictureByAddr` from `chat.members`). The clicked player is guaranteed to be in `session.chat.members` (same `PlayerIdentityData` source) — look the picture up by address here.

gonpombo8 (changes requested):
**Overall**: The infrastructure underneath this PR is solid — I went at the perf-sensitive paths hard and they held up (details at the bottom). The concern is the UI layer, where several one-off special cases sit on shared pathways and will multiply as surfaces get added. Asking for one round of consolidation before merge: the 5 blocking inline comments. The other 2 inline comments + the list below are take-it-or-backlog.

### Non-blocking: dead code + drift signals

- ~80 lines of invite-to-community plumbing across 6 files (protocol, bridge, session slice, mock, harness, tests) with zero UI consumers, whose known bugs are already documented in backlog 17b — suggest deleting and reintroducing with the feature; git history keeps the `signed()`-unwrap gotcha.
- `ReportIcon` inlined in ProfileCard while this same PR correctly adds `WalkIcon` to `design/Glyphs.tsx`.
- `.ctaSent` hand-copies the Button pill CSS instead of `<Button disabled>`.
- The address-lowercasing `nametags.tsx` isolates as `canon()` is re-inlined five times in `avatarPointer.ts`.
- `?simhover` parses the URL ad-hoc instead of joining `MockOptions`.

### Checked and fine (so nobody re-litigates)

- The `hoverPos` external store / `useSyncExternalStore` design is correct: the position must come from the engine (native overlay never delivers DOM mousemove over the world), and the store keeps a 60–120Hz stream from re-rendering the HUD tree. Verified negligible cost.
- The `getAvatarModifiers` poll can't race itself into stale state — the engine path is strictly FIFO and the click-time recheck covers the poll window. (The *style* is the frame-logic comment; the mechanism is sound. A `getAvatarModifierStream` engine-side would delete the poll entirely — backlog.)
- Pointer handlers on avatars are correctly removed via the `registered`/`present` diff within one poll cycle; the self-avatar filter has no spawn-window gap (guard and scan read the same ECS component in the same tick).
- Avatar hover events don't add HUD re-render load — avatars already emitted hover enter/exits pre-PR through the same path.

react-web/src/features/session/useEngineSession.ts
pabloes: Thanks for the detailed catch — agree the mechanism-level framing is the right end-state, but I'd argue this is **backlog, not blocking**. I've written it up under backlog item **#10** (modal/scrim consolidation), sub-point **10b**. Reasoning:

**Point 1 — reopen the same avatar, tooltip only returns after a move.** Expected, native tooltip behavior. Hover tooltips are `mousemove`/`mouseenter`-driven; dismissing a popover with a scrim click generates no new enter edge, so the tooltip returns on the next pointer move — same as closing a context menu over a button in any OS/desktop UI. Cosmetic, self-heals on a 1px move.

**Point 2 — chat/friends cards leave a frozen tooltip.** Technically true that those paths don't call `setHover([])` and `<Pointer>` stays mounted (only gated on `pageOpen`, and `friends` isn't in it) — but in practice it can't paint a stuck tooltip. `session.hover` is only populated by world-entity enter/exit edges, and the chat/friends triggers are DOM panels above the engine canvas. Reaching them with the cursor requires crossing free canvas first, which fires the hover-exit and clears `session.hover` on the way. The avatar path is the only one where the scrim drops directly onto the hovered entity with no exit — which is exactly why it's the one path that gets the explicit clear.

**Why not a small fix now.** I prototyped the minimal version (drop `setHover([])`, mask `<Pointer>`'s hints while `worldCard` is open). ~3 lines, but not a clean win: the raycast is frozen behind the scrim, so on close the stored hover/pos are stale — closing *away* from the avatar can flash a mispositioned "Show Profile" for a frame until the next move, trading "nothing until move" for "possibly-stale until move." It also just relocates the special case (handler → an App-level `worldCard ? [] : hover` ternary) instead of removing it.

**The real mechanism-level fix** is what you describe: a single scrim/Modal primitive that publishes an "overlay open" signal, with `<Pointer>` gating its hover hints on it — then *every* popup (present/future, world- or DOM-triggered) suppresses the world tooltip for free, no per-message clears. Today that can't be done cleanly: there are three separate scrim implementations (`Modal.backdrop`, `ProfileCard.scrim`, `CommunityCreateModal.scrim`) and no shared open-state to consume. Unifying them + exposing that signal is exactly the surface of backlog #10, so I'd rather land the gate there as one coherent change.

Proposing to move this to #10 and keep the current per-message clear as-is for now. Backlog #10b spells out the three sub-steps (unify scrims → expose "overlay open" signal → gate `<Pointer>`). Shout if you'd rather block on it.

react-web/bridge-scene/src/domains/avatarPointer.ts
pabloes: Done, reworked to mirror the old bevy-ui-scene avatar-tracker, which already ran this as two independent engine.addSystem timers. All three points addressed:

frame++ % 30 -> dt accumulator. Cadence is now frame-rate independent (ctx.push already passed dt; it was just ignored).
Split the two cadences. Now two separate ctx.push systems: the privacy-modifier RPC on its own slow beat (1.0s), the local entity rescan on a faster one (0.5s).
In-flight guard. The modifier poll is wrapped in a single-flight guard, so a slow engine can't stack overlapping RPCs. That makes the "only one at a time" guarantee local to this file instead of relying on the engine channel being FIFO.
Extracted the reusable bits to `bridge-scene/src/system-helpers.ts`:

```
export function throttleByDt(interval: number, fn: () => void): (dt: number) => void { ... }
export function singleFlight(fn: () => Promise<void>): () => void { ... }
```
so avatarPointer reads declaratively, with no loose counter/flag state:

```
ctx.push(throttleByDt(1.0, singleFlight(async () => { /* poll modifiers */ })))
ctx.push(throttleByDt(0.5, () => { /* rescan + attach/detach */ }))
```
Also took your new Set(...) suggestion for passportDisabled — rebuilt in one assignment (atomic by construction), which reads cleaner than clear() + refill here.

**One note**: that dt-accumulator is re-implemented inline across most bridge domains (chat, world, friends, project, nametags, avatarPreview). Migrating them to throttleByDt is tracked as backlog #21 — kept out of this PR to stay surgical.

Commit: e86d3a79.

react-web/src/features/chat/ProfileCard.tsx
pabloes: This was a scoped usability fix, noted in the backlog, to keep pushing on usability bugs and stay focused on the "nearby profile" PR for now.

But I'm now taking on the full backlog item (#7 — imperative dialog/popup helpers) instead: the same pattern we used in bevy-ui-scene (a stackable, imperative popup layer — components open popups, popups open popups, no prop-threading), adapted to React with a central Context + portals rather than a Redux store/service. useConfirm becomes sugar over a general useDialog, and the block confirm moves onto it. (The broader "consolidate the remaining bespoke scrims onto this" is the related #10.)

(work in progress…)

pabloes: The `display:'none'` card-hiding **and** its companion `r.width === 0` clamp-guard are both gone. Two steps:

1. A destructive confirm now **closes the profile card** and shows the confirm as a separate top-level dialog (Cancel / dismiss / confirm all resolve it; the card doesn't reappear / reopen via the avatar/name). With no card underneath, there's no z-stacking to work around.

2. Introduced an **imperative, stackable popup layer** (`src/design/popups.tsx`): `openPopup` / `showDialog` / `showConfirm` as plain functions over a module store, rendered by a single `<PopupHost/>`. Same *pattern* as bevy-ui-scene's popup-stack (**open a popup from anywhere, popups can open popups, no prop-threading**) but React-native, a module store read via `useSyncExternalStore`, no Redux. `showConfirm` returns a Promise, so Block is `if (await showConfirm(...)) friends.act('block', addr)`.

This also drops the `onBlock` prop-threading and App's `blockTarget`/inline-`ModalShell` state, and gives one home for confirm/alert/custom dialogs going forward (`showDialog` takes arbitrary actions). Escape/scrim close is inherited from `ModalShell`. Backlog #7 (imperative dialog helpers) is now mostly done off this.

Commit: `96f78bdd`.

react-web/src/features/chat/ProfileCard.tsx
pabloes: Problem gone by resolving other comments:
https://github.com/decentraland/bevy-explorer/pull/915#discussion_r3529180262

react-web/src/features/pointer/Pointer.tsx
pabloes: **TLDR; same behaviour as unity-explorer and old bevy-ui-scene, for backlog**

Confirmed the clip (cursor at the right edge and the right-middle prompt's label cuts off), so the "no viewport clamping" is real. Two notes though:

Small correction: there's no `overflow: hidden` on the pointer root (or `.hintsRadial`). The clip is just the viewport edge, not an overflow rule. 

I'd lean non-blocking (as you tagged it), and **it's parity with the reference: unity-explorer's** `ShowHoverFeedbackSystem` + `HoverCanvas` position a fixed layout (`CURSOR_LAYOUTS`) at the cursor with no edge-clamp or flip either (only `text-overflow: ellipsis` on the label). And since **the prompt is anchored to the cursor (not a static tooltip)**, at the extreme edge some overflow is inherent, the cursor is already sitting there. So a clamp would be an enhancement over the canonical behavior rather than a parity fix.
Noted it in the backlog as a point to review (clamp the container by its measured bounds like ProfileCard, or flip the slot sides near the edge). Happy to add it in this PR if you want it, otherwise I'll leave it as a follow-up.

react-web/src/engine/protocol.ts
pabloes: **I moved the ownership of ~~cursor~~ tooltip position to the DOM/React;** 
Engine had frame delay and was feeling slow
therefiore removed al x,y from those messages, `hoverPos` message is removed.

react-web/src/App.tsx
pabloes: fixed

PR #918, opened 2026-07-02
Particles

### LINKED ISSUE #913: Particle system
Implement https://github.com/decentraland/protocol/pull/382

hukasu:
Closes #913 

# TODO
- [x] Cube shape
- [x] Validate Cone shape
- [x] Bursts
  - [x] Bursts time
  - [x] Bursts probability
  - [ ] Looping bursts
- [x] Initial rotation
- [x] Rotation over time
- [x] Face direction of velocity
- [x] Playback state
- [x] Loop
- [ ] Prewarm
- [x] Fix local simulation space breaking billboarding
- [ ] Fix local simulation space breaking aceleration
- [ ] Fix interaction between faceTravelDirection and billboard
- [ ] Fix rotation when billboard is on
- [ ] Fix rotation when faceTravelDirection is on
- [x] Fix delayed burst playing 1 time too few


Cargo.lock
```diff
@@ -11416,7 +11455,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c5f7c95348f20c1c913d72157b3c6dee6ea3e30b3d19502c5a7f6d3f160dacbf"
 dependencies = [
  "cc",
- "windows-targets 0.52.6",
```
robtfm, line 11419: need to revert these unexpected changes (itertools, windows-*)

crates/particle_system/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "particle_system"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bevy = { workspace = true }
+bevy_hanabi = { workspace = true, features = ["2d", "3d", "gpu_tests"] }
```
robtfm: probably don't need the tests?

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+        );
+    }
+
+    if let Some(bursts_config) = &particle_system.bursts {
+        let bursts = &bursts_config.values;
+        let children = if let Some(children) = maybe_children {
+            let mut filtered_children = children
+                .into_iter()
+                .copied()
+                .filter(|child| particle_systems.contains(*child))
```
robtfm: i think this catches/replaces scene entity children with sdk `ParticleSystem` components and doesn't catch previous burst children (because they don't get `ParticleSystem`). think you need a dedicated `BurstChild` component.

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    let mut module = writer.finish();
+    if render_texture.is_some() {
+        module.add_texture_slot("color");
+    }
+
+    match playback_state {
+        PlaybackState::PsPlaying => (),
+        PlaybackState::PsPaused | PlaybackState::PsStopped => {
+            spawner_settings.with_starts_active(false);
```
robtfm: this doesn't modify the original `spawner_settings`

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+            effect_spawner.active = active;
+            true
+        } else {
+            false
+        }
+    };
+
+    for (entity, scene_entity, maybe_children) in particle_systems {
+        let active = active_scenes.contains(&scene_entity.root);
+        if !set_active(entity, active) {
```
robtfm, line 143: shouldn't this use the particle system setting as well?

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+) {
+    let mut effect_material = EffectMaterial { images: vec![] };
+
+    let max_particles = particle_system.max_particles.unwrap_or(1000);
+    let lifetime = particle_system.lifetime.unwrap_or(5.);
+    let gravity = particle_system.gravity.unwrap_or(0.);
+    let additional_force = particle_system
+        .additional_force
+        .as_ref()
+        .map(Vector3::abs_vec_to_vec3)
```
robtfm:
```suggest
        .map(Vector3::world_vec_to_vec3)
```

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+        }
+    }
+}
+
+fn particle_system_on_remove(trigger: Trigger<OnRemove, ParticleSystem>, mut commands: Commands) {
+    // On replace ParticleEffect will be replaced with a new value
+    // On despawn ParticleEffect will cease to exist anyways
+    commands
+        .entity(trigger.target())
+        .try_remove::<ParticleEffect>();
```
robtfm: also remove burst children

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,618 @@
...
+            SpawnerSettings::rate(rate.into())
+                .with_starts_active(active)
+                .with_cycle_count(!r#loop as u32),
+            &mut texture_resolver,
+            renderer_scene_context,
+            &mut effect_assets,
+        );
+    }
+
+    if let Some(bursts_config) = &particle_system.bursts {
```
robtfm, line 204: we should despawn old bursts here if bursts is None

crates/particle_system/src/plugin/mod.rs
```diff
@@ -0,0 +1,618 @@
...
+                .collect()
+        };
+
+        for (burst, entity) in bursts.iter().zip(children) {
+            debug!("Creating burst particle system");
+            make_particle_system(
+                &mut commands,
+                entity,
+                particle_system,
+                SpawnerSettings::new(
```
robtfm, line 243: we hit an assert! in release mode here when `interval: 0, cycles: 10` - needs some input sanitization. 

there are 3 asserts in the hanabi constructor, i think clamping the interval to (cycles == 1 ? 0 : 0.01, 1e9) covers them all

robtfm (approved):
2 more minor changes then please go ahead and merge

crates/particle_system/src/plugin/mod.rs
hukasu: can you check if my solution for this is satisfactory?

robtfm: It misses the is_finite assert. Not sure if we sanitize crdt f32s before here, I don’t think we do, so scene putting nan / +inf would probably still crash

PR #919, opened 2026-07-02
Pulse transport: migrate avatar movement comms off LiveKit

robtfm:
Migrates realtime avatar movement — position, velocity, rotation, animation state, emotes, and profile-version announcements — from the LiveKit data channel onto the server-authoritative **Pulse** transport (UDP/ENet natively, WebTransport on wasm; server-quantized deltas over a per-subject sliding-window baseline). The rest of the client is untouched: every Pulse `PlayerState` is reconstructed into an `rfc4::Movement` and pushed through the existing `global_crdt` / `foreign_dynamics` pipeline, so nothing downstream learns Pulse exists.

## Highlights
- **Native ENet driver** (`rusty_enet`) with the full connection lifecycle: signed-auth handshake, teleport, and reconnect.
- **WebTransport driver for wasm** (`crates/comms/src/pulse/wasm.rs`) — the browser speaks Pulse over WebTransport (HTTP/3 / QUIC): a length-framed reliable bidi stream plus sequenced/unsequenced datagrams, matching the server's framing exactly. Native and wasm share the byte-boundary `PulseDriverChannels` seam; only the driver differs.
- **Proto-driven quantization ABI** — the dequantize/quantize accessors are generated from the pulse `.proto` field options (`build_quant.rs`), so the client can't silently drift from the C# server's `protoc-gen-bitwise` constants. Tracks protocol #431 (power-law velocity) and the quantized `PlayerState`.
- **Foreign-avatar smoothing** — position deltas are treated as a ±half-step *box* and dead-reckoned within it rather than snapped to the quantized centre, killing low-speed jitter. Head yaw/pitch preserve sign across the unsigned `[0,360]` quantizer.
- **Scene-driven animation decoupled** from movement (rides its own packet; animation stays on LiveKit).
- **Pulse gated to the realm island** — scene rooms stay on LiveKit; no double-delivery.

## ⚠️ Do not merge until the Pulse WebTransport server is deployed
Both drivers now exist — native (UDP/ENet, port 7777) and wasm (WebTransport, port 7443). The remaining gate is server-side: the Pulse server's WebTransport endpoint must be deployed to the target realms before the web client can connect. Until then the wasm build's Pulse transport won't establish (the default targets `pulse-server.decentraland.org:7443`), so a web client on a Pulse realm has no working movement transport. Draft until the server's deployed.


eordano:
Heads up: C#'s `Math.Round` defaults to banker's rounding, so the `decentraland/Pulse` server and this PR will be diverging in some very corner cases: `build_quant.rs` uses `f32::round`

robtfm:
The server is receiving/rebroadcasting quantised data now, so bevy/bevy (and bevy/godot) will agree. Bevy/unity may see mismatches when exactly on the rounding threshold … I’d prefer to ignore that rather than try to match I think …

PR #924, opened 2026-07-03
CI: cheaper dependabot runs, cached wasm builds, faster main deploys

gonpombo8:
## Summary

Three changes to make CI cheaper and deploys faster.

**Dependabot was burning our quota**

Every dependabot PR ran the full CI matrix. macOS runners bill at 10x, so a single dependency bump cost around 700 billed minutes. Now dependabot PRs skip the test matrix and the scene tests. The web build still runs on them, because it's the only job that compiles the npm packages, so a broken npm bump can't merge green. There's also a new `dependabot.yml` that groups everything into one weekly PR per ecosystem instead of one PR per package.

**The web deploy was rebuilding everything, every time**

The deploy job took ~30 minutes, and almost all of it was avoidable:

- 15 min compiling the wasm engine from scratch (the job had no cache at all)
- 9 min of wasm-opt, even on PRs, whose binary is never published anywhere
- 1 min compiling wasm-pack from source instead of just downloading it

The interesting part is the cache. Some jobs already used rust-cache, but it barely helped, and the reason took a while to find. GitHub gives a repo 10 GB of cache space. Every PR run was saving its own ~2 GB copy of the build cache, and a cache saved by a PR can only be restored by that same PR. Other PRs and main can't read it. So the quota was full of copies nobody could reuse, and GitHub kept evicting the caches that actually mattered to make room. That's why even the jobs with caching kept running cold.

The fix: caches are now only saved from main (`save-if`). PRs restore main's cache and never write their own. This works because GitHub lets a PR read caches from its base branch, so every PR starts from the latest main build and only recompiles what it changed. It also means a PR can never write a cache that main reads, so there's no way to poison the deploy build from a PR.

And it's safe to restore a stale cache: it's just a warm `target/` folder. Cargo checks every crate's fingerprint against the real sources and lockfile and rebuilds whatever doesn't match. A stale cache costs minutes, never correctness.

Smaller things in the same area: wasm-opt is skipped on PRs (`--no-opt`) since only main publishes; wasm-pack is a pinned binary download now; the Check job is gone (clippy compiles the same code and lints on top of it); rustfmt lost a pile of apt installs it never needed.

Expected numbers: main deploys ~15 min (wasm-opt is the part we can't skip), PR deploy builds ~6-8 min once main has saved a cache. The first run after merging is still cold.

**Main pushes were queueing behind each other**

All main runs shared one concurrency group, so a new push couldn't start until the previous run's slowest job finished. The #925 deploy sat 26 minutes doing nothing, waiting for #901's Windows tests. Now every main run starts immediately. If a newer push lands while tests are still running, the older test jobs get cancelled (only the newest matters). The deploy job keeps its own group so two pushes can't publish at the same time.

## What could break

- Dependabot PRs don't run the test matrix. A cargo bump that compiles but fails tests would only surface on the main push after merge.
- On quick back-to-back merges, the older push's test jobs show as cancelled. Only the newest one finishes.
- The Check status disappears from PRs. Nothing depends on it (no branch protection).
- PR wasm binaries are bigger since they skip wasm-opt. The published artifact from main is unchanged.
- If main gets no pushes for 7 days the cache expires and the next run is cold. It re-seeds itself on that run.

## How to test

- Merge and watch the main run: the deploy should take ~15 min and save the `wasm-web` cache.
- Open any PR after that: the deploy build should restore it and finish in ~6-8 min, with `--no-opt` visible in the wasm-pack log.
- Next dependabot PR: matrix jobs skipped, web build runs.
- Push twice to main a few minutes apart: the second run starts right away and the first run's test jobs cancel.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,42 @@
+# Keep dependabot cheap: one grouped PR per ecosystem per week (each PR run of the full CI
+# matrix bills ~700 runner-minutes; the heavy jobs are also gated off dependabot in ci.yml).
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 2
+    groups:
+      cargo-deps:
+        patterns: ['*']
+
+  - package-ecosystem: npm
+    directory: /react-web
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 2
+    groups:
+      react-web-deps:
+        patterns: ['*']
+
+  - package-ecosystem: npm
+    directory: /react-web/bridge-scene
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 2
+    groups:
+      bridge-scene-deps:
+        patterns: ['*']
+
+  - package-ecosystem: npm
+    directory: /deploy/web
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 1
+
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    open-pull-requests-limit: 1
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,9 +34,15 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 1
+    groups:
+      deploy-web-deps:
+        patterns: ['*']
 
   - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: monthly
     open-pull-requests-limit: 1
+    groups:
+      actions-deps:
+        patterns: ['*']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,36 +11,11 @@ concurrency:
 name: Continuous integration
 
 jobs:
-  check:
-    name: Check
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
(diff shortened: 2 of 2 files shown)


robtfm:
Pushed 6e8a578a on top of this: instead of gating the heavy jobs purely on the dependabot actor, a small `paths-filter` job now detects whether the rust dependency graph changed (`Cargo.lock`, any `Cargo.toml`, `.cargo/config.toml`) and coverage scales on that.

Rationale: the actor-only gate had two coverage gaps. A cargo bump merged green with zero Windows/macOS compile coverage (clippy is linux-only), and the "failures surface on the main push" safety net silently disappeared if a merge push ever had `actor == dependabot[bot]` (e.g. `@dependabot merge`). Tying the expensive legs to dep-file changes fixes both, and also buys back minutes on ordinary source PRs — the matrix's main value is catching platform-specific dependency breakage.

Resulting behavior:

| change | Test Suite | Scene tests |
|---|---|---|
| human PR/push, source only | ubuntu only | skipped |
| human PR/push touching dep files | full 3-OS matrix | runs |
| dependabot cargo bump (PR) | ubuntu only | skipped |
| dependabot npm/actions bump (PR) | skipped | skipped |
| merge of a cargo bump to main | full matrix (human actor + lock diff) | runs |

So cargo bumps now get a 1x-billed unit-test run at PR time and the full matrix post-merge, while npm/actions bumps stay as cheap as before (they can't affect the rust build). `paths-filter` is SHA-pinned to v3.0.2 and works on both events — PR file list via the API, push diffed against the pre-push commit (hence its checkout step).

PR #942, opened 2026-07-06
Omit null fields from deployed profile JSON

gonpombo8:
## Summary

Profiles deployed by bevy-explorer serialize every `Option::None` field as an explicit `"field": null` (`nameColor`, `snapshots`, `avatar.name`, …), while other clients omit absent fields. Present-but-null fields can fail to parse in other explorers, leaving the profile unusable there until a clean entity is re-deployed. This started affecting accounts when #877 added the `nameColor` field: since then every re-deploy (e.g. any backpack change) writes `"nameColor": null`. Verified against a real broken profile entity on the catalyst.

Changes:
- `skip_serializing_if = "Option::is_none"` on all optional profile wire-format fields, so absent fields are omitted instead of null
- `process_profile`: a `setAvatar` base that omits eyes/hair/skin colors now preserves the existing values instead of wiping them

## What could break

- Any consumer that distinguished `"field": null` from an absent field in bevy-deployed profiles or comms profile messages. None known — JS sees `undefined` either way, serde deserializes missing `Option` as `None`, and omitted fields match what other clients already deploy.
- Colors can no longer be *cleared* via `setAvatar` base — intentional, they were never meaningfully clearable.

## How to test

1. Log in with a web3 account, change your look in the backpack, close it (triggers a re-deploy)
2. Fetch the deployed entity: `curl -s https://peer.decentraland.org/content/entities/active -X POST -H 'content-type: application/json' -d '{"pointers":["<address>"]}'` — verify the avatar JSON contains no `null` values and no `nameColor`/`snapshots` keys
3. Enter Decentraland with another explorer client using the same account — profile loads correctly

### DIFF
--- a/crates/common/src/profile.rs
+++ b/crates/common/src/profile.rs
@@ -34,17 +34,27 @@ impl AvatarColor {
     }
 }
 
+// Optional fields are omitted when absent, not serialized as `null`: other explorers
+// fail to parse present-but-null fields in deployed profiles.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub struct AvatarWireFormat {
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub body_shape: Option<String>,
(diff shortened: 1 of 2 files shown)


robtfm:
happy to merge it.

either the other clients or the server should be robust against bad data - more clients are popping up every day

PR #953, opened 2026-07-09
Fix web preview launch, mock SSO leak, and world-name URL echo

gonpombo8:
## Summary

Fixes for the react-web launch flow, found while debugging why scene preview (`sdk-commands start` → `?preview=true&realm=http://127.0.0.1:8000`) landed on the Places list instead of the preview scene:

- **Every `?realm`/`?position` launch goes straight in — never the Places picker.** The `/about` probe still runs before launch, but its outcome is explicit: **ok** launches, **404** shows the World-not-found modal, and **anything else** (network/permission failure, or a 4s timeout — the probe used to hang forever on a pending browser local-network prompt) shows the not-reachable modal. Previously the picker could reappear behind the error, and a hung probe stranded the loading screen silently.
- **Non-hex SSO identities are ignored by the stored-login scan.** The mock bridge (`?mock=1&previousLogin=1`) seeds `single-sign-on-0xmock…` into per-origin localStorage; real engine sessions on the same origin picked it up, showed "Welcome back!", and `/login_identity` failed after launch with `bad root address: Invalid character 'm' at position 0`. The scan now requires a real `0x`+40-hex key and SIGNER payload, mirroring the engine's parse (login.rs).
- **World URLs keep their short name.** Entering `realm=boedo.dcl.eth` rewrote the param to the expanded `worlds-content-server…/world/…` form via the engine's URL sync; boot.js now reverses the `map_realm_name` expansion when writing the param.

## What could break

- Non-404 HTTP errors from `/about` (e.g. a 500) now read as "isn't reachable" instead of "doesn't exist".
- The probe adds up to 4s before the error surfaces when the target hangs (timeout races the fetch); success is as fast as before.
- The boot.js prefix strip assumes the production worlds-content-server host; worlds served from other hosts keep their full URL in the param (same as before).

## How to test

1. `cd react-web && npx vitest run` — 204 tests, including: probe ok launches with no picker, 404 shows World-not-found, unreachable realm shows not-reachable without launching.
2. Preview: run a scene (`npm run start` in any scene folder), then open `http://localhost:5173/?preview=true&realm=http://127.0.0.1:8000&position=0,0` (react-web `npm run dev`) — after Jump in it should enter the scene directly, no Places list. Stop the scene server and retry — the not-reachable modal should appear instead of a broken launch.
3. Mock leak: visit `?mock=1&previousLogin=1` once, then reload without `?mock` — the login screen should offer sign-in/guest, not a broken "Welcome back!".
4. World name: open `?realm=boedo.dcl.eth`, enter, and confirm the address bar still shows `realm=boedo.dcl.eth`; `?realm=nope.dcl.eth` should show the World-not-found modal.

Note: testers running from a locally built engine need a wasm built after #928 (`wasm-pack build --target web …`) — older local builds call the pre-#928 `set_url_params` convention and scramble the URL params.

### DIFF
--- a/deploy/web/engine/boot.js
+++ b/deploy/web/engine/boot.js
@@ -149,8 +149,13 @@ const config = window.__bevyBootConfig ?? {}
 // the same place. Defaults are omitted so the canonical entry URL stays clean.
 const DEFAULT_SERVER = 'https://realm-provider-ea.decentraland.org/main'
 const DEFAULT_PORTABLES = 'basiccontroller.dcl.eth'
+// The engine connects to the EXPANDED world url (ipfs map_realm_name turns `name.dcl.eth` into
+// worlds-content-server…/world/name.dcl.eth) and echoes that back here. Reverse it so the address
+// bar keeps the short name the user typed; a reload re-expands it the same way.
+const WORLDS_PREFIX = 'https://worlds-content-server.decentraland.org/world/'
 window.set_url_params = (position, server, system_scene, portables, preview) => {
   try {
+    if (server.startsWith(WORLDS_PREFIX)) server = server.slice(WORLDS_PREFIX.length)
     const urlParams = new URLSearchParams(window.location.search)
     // absent when the realm won't honour a position (worlds spawn at their base scene)
     if (position != null) urlParams.set('position', position)
--- a/react-web/src/engine/mockBridge.ts
+++ b/react-web/src/engine/mockBridge.ts
@@ -138,6 +138,9 @@ const DEFAULTS: MockOptions = {
 
 // In ?mock=1&previousLogin=1, seed a fake same-domain SSO identity so the session's
 // localStorage read lights up the "welcome back" reuse flow (no real auth site in mock).
+// The 0xmock… address is DELIBERATELY not hex: localStorage is shared with real engine
+// sessions on this origin, and sso.ts ignores non-hex identities — a valid-looking address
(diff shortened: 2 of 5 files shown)


regenesis-claw (approved):
Solid set of fixes — clean code, thorough tests, and the PR description is excellent.

**Security scan (no issues found):**
- `isHexAddress` regex is correctly anchored (`^0x[0-9a-fA-F]{40}$`) — no ReDoS risk, no bypass via padding/injection.
- `isLocalRealm` properly uses `new URL()` for hostname parsing (avoids string-matching pitfalls like `localhost.evil.com`). Coverage of `localhost`, `127.0.0.1`, `[::1]`, and `0.0.0.0` is comprehensive for dev scenarios.
- The `WORLDS_PREFIX` strip in boot.js is safe — `startsWith` + `slice` can't produce injection or redirect since the prefix is a hardcoded constant.
- SSO identity validation (`SIGNER` payload hex check) correctly mirrors the engine's `login.rs` parse, preventing the mock address leak without breaking real auth chains.

**Code review (all clear):**
- The local-realm skip is well-scoped: only skips the `/about` probe (the engine still validates connectivity on its own retry path), so the security posture for remote realms is unchanged.
- Tests cover the three key paths (local launch, remote 200, remote 404) with proper spy cleanup in `finally` blocks.
- Comments are unusually good — they explain *why*, not just *what*, which makes future maintenance much easier.

Reviewed by Claw 🤖

robtfm (approved):

PR #954, opened 2026-07-09
Run the react-web HUD on native via CEF offscreen rendering

robtfm:
Builds on #912's engine plumbing (SystemApi BridgeToPage/GetBridgeStream + the BroadcastChannel polyfill) and replaces the macOS-only wry/NSPanel overlay with cross-platform CEF offscreen rendering composited by the engine.

## What's here

- **`crates/cef_offscreen`** — offscreen CEF webviews for bevy 0.16 (extracted from not-elm/bevy_cef v0.1.0, MIT/Apache attribution retained), with fixes: renderer V8 context leak, cursor-icon entity spray, cef:// scheme flags (ES modules), query stripping, per-app cache path, sibling render-process helper (`decentra-bevy-cef`, orphan watchdog — no dangling processes), browser calls marshalled onto the CEF UI thread for the multi-threaded message loop (Windows/Linux).
- **HUD compositing** — CEF paints the react-web page into a fullscreen `ImageNode`; per-pixel alpha at the cursor gates engine world input (`FocusPolicy` flips with it, so scene UIs underneath stay live); all keys/mouse forward to the page (web DOM parity: outside-click blurs text fields, page hotkeys work). Proper RAWKEYDOWN/CHAR/KEYUP synthesis, XKB keycodes on Linux, wheel line-delta scaling, multi-click selection, mac editing shortcuts (frame edit commands + DevTools editor commands). Text focus is relayed both ways via InputPriorities: HUD fields claim the engine keyboard, and engine/scene text fields set `window.__engineTextFocus` so typed letters don't fire HUD shortcuts.
- **Native sign-in** — "Start with account" runs the engine's remote-wallet flow (auth site opens in the external browser; the verification code streams to the page as a `loginCode` message, with cancel support). Web keeps its same-origin SSO redirect — `LoginDriver.loginNew` presence is the switch.
- **Files only, single process** — the page ships in `assets/react-hud` over the `cef://` scheme; the bridge-scene loads as a **file realm** (`file://` support in IpfsIo, with reads bounded to canonicalized user-registered roots — urn-smuggled `file://` baseUrls are refused); no local servers. `--server` injects `?realm=` so the HUD skips its places picker; destination picks map to changeRealm/teleport on the native driver. Until the bridge-scene connects, a built-in boot shim answers the login rpcs and the page assumes "loading".
- **Dev/build** — `justfile` recipes (wasm / native-debug / native-release, staleness-checked react bundle), `react-hud-cef` as a default cargo feature (disabled in test mode), readme build steps, packaging + CI updates for the helper binary and CEF framework/resources per platform (AppImage bundles assets + CEF, excludes libxkbcommon so the host supplies its matched pair).

Tested on **macOS**, **Linux** (source build and packaged AppImage, including NixOS) and **Windows**. Linux needed its own round of fixes along the way: single-threaded zygote (no watchdog thread there), in-process stub GPU service, XKB keycodes, and the libxkbcommon AppImage exclusion. The macOS release-workflow leg is temporarily disabled pending the org's Apple notarization secrets (commented in package.yml).


crates/cef_offscreen/src/core/browser_process/browser_process_handler.rs
```diff
@@ -0,0 +1,63 @@
...
+        let Some(command_line) = command_line else {
+            return;
+        };
+
+        command_line.append_switch(Some(&"disable-web-security".into()));
+        command_line.append_switch(Some(&"allow-running-insecure-content".into()));
+        command_line.append_switch(Some(&"disable-session-crashed-bubble".into()));
+        command_line.append_switch(Some(&"ignore-certificate-errors".into()));
+        command_line.append_switch(Some(&"ignore-ssl-errors".into()));
+        command_line.append_switch(Some(&"enable-logging=stderr".into()));
```
regenesis-claw, line 59: **Security**: `ignore-certificate-errors` + `ignore-ssl-errors` bypass TLS validation in all CEF child processes. If the HUD page makes any HTTPS fetches (catalyst APIs, CDN images, etc.) this exposes them to MITM. Are these needed? The parent process's `disable-web-security` handles the CORS issue; these appear to be a separate concern (cert validation, mixed content). If all HUD content is local (`cef://`), these are safe to remove.

crates/ipfs/src/lib.rs
```diff
@@ -1582,6 +1599,45 @@ impl AssetReader for IpfsIo {
...
+                            .iter()
+                            .any(|root| canonical.starts_with(root))
+                    })
+                    .ok_or_else(|| {
+                        warn!("refusing file read outside registered file realms: `{remote}`");
+                        AssetReaderError::Io(Arc::new(std::io::Error::other(format!(
+                            "file realm read `{remote}`: not found or not permitted"
+                        ))))
+                    });
+                let data = ipfs_io_read_state.send_failure(local.and_then(|local| {
```
regenesis-claw, line 1630: 👍 Clean path-traversal prevention: canonicalize + allowed-root check. The fact that `allowed_file_roots` is only populated from `lookup_local_realm` (which requires `parent_scene.is_none()`) means scene-supplied URNs can't sneak in `file://` baseUrls.

regenesis-claw (commented):
## Security-focused review (react-hud-cef)

This is a massive PR (~8k additions) — I focused on the security-sensitive areas: file:// path handling, CEF flags, IPC bridge, input routing, and the `cef://` scheme handler. I skimmed the CI/packaging, keyboard translation, and CEF boilerplate.

### ✅ Solid security decisions

1. **File-realm path traversal** — `IpfsIo::read` canonicalizes `file://` paths and checks them against `allowed_file_roots`. The `lookup_local_realm` gate (`parent_scene.is_none()`) correctly prevents scene-initiated file reads. URN-smuggled `file://` baseUrls from remote realms/portables are properly refused.

2. **`cef://` scheme handler** — query/fragment is stripped before asset lookup (`split(['?', '#'])`) preventing path confusion. The assets are served from the bevy asset pipeline, not raw filesystem.

3. **BroadcastChannel polyfill scope** — defined globally in `init.js` but guards on op existence (`op_bridge_to_page`, `op_get_bridge_stream`), so non-super-user scenes get an inert channel. Safe.

4. **Input isolation** — `InputPriority::TextEntry` reservation correctly suppresses engine key actions during HUD text editing. Bidirectional: engine text fields also signal the page via `__engineTextFocus`.

### ⚠️ Items worth discussing

1. **CEF child-process SSL bypass** (`browser_process_handler.rs:55-58`) — the child process launch adds `ignore-certificate-errors`, `ignore-ssl-errors`, and `allow-running-insecure-content`. The browser process already has `disable-web-security` for CORS reasons (understood — `cef://` origin has no CORS allowlist). But bypassing certificate validation in child processes means if the HUD ever fetches remote HTTPS resources (API calls, CDN assets), MITM attacks become trivial. If the HUD is truly local-only via `cef://`, these flags are dead code and should be safe to remove. If it does make remote calls, this is a real risk.

2. **Full-screen CPU texture copy** — `RenderAssetUsages::all()` keeps a CPU copy of every HUD paint for the alpha test in `hud_alpha_at`. At 1080p that's ~8 MB/frame continuously. This is by design (needed for per-pixel passthrough), but worth being aware of as resolution scales up.

3. **Mouse wheel leaks to engine** (acknowledged in header comment, line 36 of `react_hud_cef.rs`) — wheel events over the HUD still reach the engine. Could cause zoom while scrolling a HUD list.

4. **`visibilityState` override** (`main.tsx`) — `Object.defineProperty(document, 'visibilityState', ...)` is a clever workaround for CEF throttling but makes the page unable to ever detect actual background state. Fine for this use case, but something to remember if background detection is ever needed.

### Nits (non-blocking)

- The `BroadcastChannel` polyfill only supports a single `onmessage` listener (overwrites on repeat `addEventListener`). This matches current usage but would break if any code adds multiple listeners.
- `MultiClick` detection uses `time.elapsed_secs_f64()` which resets on app pause — unlikely to matter in practice.

Overall the security model for file reads is well-designed. The main actionable item is the SSL bypass flags in child processes — recommend removing them if not needed, or adding a comment explaining what remote calls require them.

robtfm:
TLS bypass flags removed in b614a01c — the HUD does make remote https fetches (catalyst profiles/thumbnails, places), so certificate validation now stays on. `disable-web-security` is retained deliberately: it's needed for CORS, not TLS. We verified the API posture directly — catalyst lambdas send `Access-Control-Allow-Origin: *` (fine from any origin), but places.decentraland.org echoes ACAO only for allowlisted `*.decentraland.org` origins, so fetches from the `cef://localhost` origin (or any third-party web origin) are blocked without it. Removing it was tested and broke the places page exactly as predicted.

PR #991, opened 2026-07-21
fix/ui/04-outfits: backpack — outfit equip resolves by urn (off-page data loss), def cache, drop phantom head slot

pabloes:
### What
Post-merge follow-up to #984 (fix/react-web-hud-03), five commits:

- **Equipping a saved outfit is now bridge-authoritative**: the bridge resolves the outfit's wearables **by urn** and re-emits the full `wearables` equipped set after deploying (mirroring `getWearables`), replacing the lossy client-side rebuild.
- **Wearable-definition cache by item urn** in the bridge (`defByItemUrn`, session-lifetime, in-memory) — repeat resolves (backpack reopen, outfit equip) hit memory instead of the catalyst. Mirrors bevy-ui-scene's `catalystMetadataMap`.
- **Dropped the phantom "head" category slot** — rendered with a fallback (∞/ALL) glyph and always filtered to an empty grid.
- **Fresh-review fixes (5th commit)**: the authoritative `wearables` emit now also reconciles the grid page's per-item equipped flags (stale markers/slots shadowed the truth after an outfit equip), and a failed defs resolve no longer shrinks the emitted set — chunk failures log, unresolved urns stay equipped under `'unknown'` (no slot, but never dropped from a later deploy).
- **Refreshed the `backpack-wearables` visual baseline** (one tile fewer) and added backlog *20 (unify avatar-equip authority — the architecture follow-up this fix surfaced).

### Why
- **P1 (silent data loss) from the #984 post-merge review**: equipping a saved outfit whose items were off the loaded catalog page dropped them from `backpack.equipped` — the session rebuilt the set from `current catalog page ∪ previous equipped`, so off-page urns resolved to nothing. The category column immediately under-reported, and the **next single-item equip deployed without those items, actually removing them from the avatar**. Needs >16 owned wearables (PAGE_SIZE, rarity-DESC page 0) to reproduce; `?mock=1` masked it because the mock re-emits `wearables` on `equipOutfit` while the real bridge didn't. bevy-ui-scene never had this bug — its outfit equip always resolved by urn (`fetchWearablesData(...)(...wearables)`); this restores that model in the bridge.
- **Definitions are stable within a session** (name/category/thumbnail per item urn; only ownership/equipped state is live), so re-resolving them on every `getWearables`/`equipOutfit` was wasted catalyst load — and the new `equipOutfit` re-emit adds a resolve that the cache makes mostly free.
- **"head" is not an equipable category**: the schemas define it only as a hide/replace *target* (`hides: ["head"]`), nothing is published with `category: "head"`, and unity-explorer's Backpack prefab serializes exactly 18 AvatarSlots — no `head` (its `NftCategoryIcons` head glyph is for cards, the likely source of the porting confusion). There's also no `head.png` in our category icons, so the tile fell back to the ALL glyph.

### How
- **bridge `wearables.ts`**: extracted `resolveEquippedSet(urns)` — indexes item→token (`tokenUrnByItem`) and resolves defs via `resolveByUrn` — shared by `getWearables` and `equipOutfit`. `resolveByUrn` now serves cache hits from `defByItemUrn` and only fetches misses (batched, 50/req); misses are not cached, so transient failures retry.
- **bridge `outfits.ts`**: after the `setAvatar` deploy, `equipOutfit` emits `{ kind: 'wearables', equipped }` resolved from the outfit's urns.
- **session (`useEngineSession`)**: `equipOutfit` only dispatches — removed the optimistic rebuild and the now-unused `outfitsRef`. The bridge emit is the authoritative equipped set. (Single-item `equip` keeps its optimistic rebuild — its pool is lossless by construction; unifying that model is backlog *20.)
- **`BackpackPage`**: `'head'` removed from `CATEGORY_ORDER`; comment documents why it must not come back.
- **tests (`wearables.test.tsx`)**: the outfit regression now encodes the new contract — the outfit contains an item **off** the loaded page, the bridge re-emit carries it, and a later single-item equip keeps it.

### Verification
- Typecheck clean (app + bridge-scene), **257/257** unit tests (Node 24).
- `?mock=1` browser pass: outfit equip populates the category slots, column shows 18 categories, no console errors.
- Baseline: `--update-snapshots=all` was required — the one-tile diff sits **under** the full-page 1% `maxDiffPixelRatio`, so the stale baseline still passed (worth remembering for future small diffs).

### Known limitations
- **Saved Outfits still has no visual baseline** — deferred deliberately; the section's look is about to be reworked, baseline lands with that.
- **Single-item `equip` remains client-authoritative with no bridge re-emit** — a failed deploy still diverges HUD from avatar until reload. Documented with the fix path as backlog *20.
- The def cache is session-lifetime: a wearable re-published under the same urn shows stale metadata until reload (accepted trade-off; defs rarely change).


regenesis-claw (approved):
Clean, well-documented fix for a real P1 data-loss bug. Security review: no new external inputs, no auth changes, no injection surfaces — the urn cache is in-memory session-lifetime with no persistence or cross-session leakage risk. The `startsWith(\`${w.urn}:\`)` token-urn matching is safe given the structured urn format.

**Key observations (non-blocking):**

1. **Re-emit on deploy failure (outfits.ts):** `ctx.send` runs after `.catch()` settles, so a failed `setAvatar` deploy still re-emits the outfit's wearables — HUD diverges from avatar. Acknowledged in the PR body and tracked as backlog #20 (unify authority model), so not blocking. The old behavior was strictly worse (silently dropped off-page items).

2. **Cache unbounded but bounded in practice:** `defByItemUrn` has no max-size eviction, but the population is naturally bounded by unique wearable urns encountered in a session — realistic upper bound is low hundreds. No concern.

3. **Excellent backlog #20 write-up** — the architecture follow-up (bridge re-emit for every mutation, epoch sequencing, urn coverage) is thorough and actionable.

Test coverage is solid: the regression test encodes the exact off-page scenario, and the reconciliation of page flags is verified. 257/257 tests pass, CI all green. LGTM ✅

gonpombo8 (approved):

PR #992, opened 2026-07-21
fix/ui/05-emote-hotkeys: emotes — number keys play the wheel slot

pabloes:
## What

- Emote hotkeys: while the wheel is open, number keys 0-9 now play the emote in that slot (and close the wheel). A number with the wheel closed does nothing. Covers both gestures:
  - `b + number` (hold B, which opens the wheel, then press a number)
  - `b , number` (tap B to open the wheel, then press a number)
- Empty-wheel defaults: a fresh profile (no equipped emotes) now renders the 10 base emotes instead of an empty wheel, and equipping into a fresh profile keeps the other defaults instead of blanking the remaining 9 slots. Base-emote short names are normalized to their full off-chain urn so equipped base emotes match a slot (the engine stores them by short name; the catalog keys on the full urn).
- Updated the wheel hint text to match the tap-B-then-number behavior ("Press [B+num] to run an emote").

## Why / approach

The hotkey handling lives in the HUD's capture-phase `window` keydown listener (`useMenuShortcuts`, renamed `handler` to `keyDownHandler`), next to the existing letter shortcuts. It does not go through the engine's `QuickEmote` system action on purpose: that action only fires while the engine canvas holds keyboard focus, so it never arrived once the HUD (or the open wheel) took focus. A DOM window listener sees the key regardless of focus, which is exactly what the B shortcut already relies on.

Gating on "wheel open" is enough for both gestures, since B opens the wheel first in either case. Keystrokes aimed at text fields (e.g. the chat input) are still ignored.

The empty-wheel fill mirrors Unity's SelfProfile behavior: the bevy runtime does not seed the default emotes into `getPlayer().emotes`, so the HUD fills them (slot i = base emote i) when every slot is empty. It is all-or-nothing, so once any slot is equipped the remaining empties stay empty.

## Tests

- New unit test in `menuShortcuts.test.ts`: a number plays that slot only while the wheel is open; a bare number is a no-op; an empty slot is a no-op.
- Full react-web suite: 258/258 passing. Typecheck clean (react-web and bridge-scene).
- Verified in the browser with `?mock=1`: B opens the wheel, a number plays and closes it, a number with the wheel closed does nothing, no console errors.

## Note

Stacked on #991 (`fix/ui/04-outfits`); base is that branch, so this PR's diff is the emote hotkey commit plus the empty-wheel default-fill fixes and the hint-text change. Retarget to `main` once #991 merges.

### DIFF
--- a/react-web/bridge-scene/src/domains/emotes.ts
+++ b/react-web/bridge-scene/src/domains/emotes.ts
@@ -25,6 +25,16 @@ function itemUrn(urn: string): string {
 }
 const isBase = (urn: string): boolean => BASE_EMOTES.includes(itemUrn(urn))
 
+// The 10 wheel slots to show for a profile: its equipped emotes positionally, or — when the wheel is
+// entirely empty (a fresh profile) — the base emotes in order (slot i = BASE_EMOTES[i]). The bevy
+// runtime doesn't seed the defaults into getPlayer().emotes (bevy-ui-scene hits the same empty array),
+// so the HUD fills them here, mirroring Unity's SelfProfile empty-wheel fill: it's all-or-nothing, so
+// once any slot is equipped the remaining empties stay empty.
+function equippedSlots(emotes: readonly unknown[] | undefined): string[] {
+  const slots = Array.from({ length: SLOT_COUNT }, (_, i) => String((emotes ?? [])[i] ?? ''))
+  return slots.every((u) => u === '') ? [...BASE_EMOTES] : slots
+}
+
 // Title-case a base-emote id ("raisehand" → "Raise Hand"); custom emotes use the catalog name.
 function baseEmoteName(urn: string): string {
   const i = urn.indexOf('base-emotes:')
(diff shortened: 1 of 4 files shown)


regenesis-claw (approved):
Clean, focused PR — emote hotkeys, empty-wheel default fill, and URN normalization all look correct.

**Security scan** (per standing request): no concerns — key handler properly gates on `e.repeat`, chords, and text-field targets; no user-controlled strings flow into anything sensitive; no network/auth changes.

**Code notes:**
- `fullEmoteUrn` correctly distinguishes bare short names (no `:`) from full URNs — simple and robust.
- `equippedSlots` returns a fresh copy (`[...BASE_EMOTES]`) when the wheel is empty, avoiding mutation of the constant — good.
- The hotkey handler's "wheel must be open" gate elegantly covers both hold-B and tap-B gestures without extra state.
- Test covers the key paths (wheel open → plays slot, wheel closed → no-op, empty slot → no-op).
- Stacked on #991 — retarget to `main` once that merges.

LGTM ✅

PR #997, opened 2026-07-23
fix/ui/06-wgpu-crash-popup: engine-host — surface fatal wgpu error floods as a crash modal

pabloes:
## What

- Adds an **"alive but rendering-dead" flood detector** to `deploy/web/engine/boot.js` (host-integration JS): an identical engine-core `console.error` repeating **≥30× within 4s** now trips `crash(msg, 'runtime')` → the existing `EngineErrorModal`, instead of leaving the user on a silent black screen.
- **Latched** (fires once, not a stack of popups); **Dismiss** re-arms, so a persistent flood re-triggers.
- **Denylist** for benign identical repeats (sandbox `"Message too large"`).
- Host-side only (`boot.js`); **no WASM rebuild**.

## Why

During the **2026-07-17 playtesting session**, multiple users hit a **black screen / engine crash** with no error popup — just an endless console spam of a WebGPU validation error (`pbr_alpha_blend_mesh_pipeline` attachment mismatch on `main_opaque_pass_3d`). The user is stranded with no feedback and no recovery path.

No popup fired because the crash modal (`EngineErrorModal`) is fed by two signals, and the wgpu error is **neither**:

1. **Rust panic** — `boot.js` sniffs `console.error` for `"panicked at …"`.
2. **Heartbeat stall** — the watchdog fires `__onEngineCrash` only when frames **stop**.

The wgpu error takes a third path that only logs: `src/lib.rs` registers `wgpu_device().on_uncaptured_error(|e| error!("captured wgpu error: {e:?}"))` — logged and swallowed. Decisively, the **render loop keeps beating** (each frame re-sets the bad pipeline → black frame, but the heartbeat still ticks), so the stall watchdog sees a healthy engine and never fires. → endless spam + black screen + **no popup**.

The structural gap: the watchdog detects *"engine STOPPED"* but had no detector for *"engine RUNNING but rendering is dead"*.

## How

Reuses the same `console.error` patch that already sniffs panics:

- **Per-signature counters** (whitespace-normalized, otherwise **exact**) over a 4s window; ≥30 identical repeats ⇒ `crash(msg, 'runtime')` → the existing modal.
- Exact-signature keys so distinct errors (different assets/coords) never pool into a false flood; a per-signature Map so an identical per-frame error still floods even when interleaved with a second error.
- `rearm()` (modal dismissed, or frames resumed) clears the counters.
- Skips the watchdog's own `[engine error]` / `[crash watchdog]` logging so it can't self-trip. Scene (UGC) `console.error` runs in the sandbox worker's own console and never reaches this main-thread patch, so the stream stays engine-core.

## Testing

Verified **end-to-end against the real engine build** (react-web :5199, real engine, not `?mock=1`) via a synthetic trigger — spamming an identical `console.error` (incl. text sourced from a **real** `GPUValidationError`) 35× on the main thread:

- Flood → **"The world crashed"** runtime modal with the wgpu error text + Dismiss/Reload. ✅
- Fires **once** (latched), not a stack of popups. ✅
- **Dismiss** re-arms; a second flood re-triggers (matches persistent-error behaviour). ✅
- **Denylist**: 50× `"Message too large"` does **not** trigger. ✅
- A `throw` does **not** trigger it (different channel — uncaught exceptions don't hit the patched `console.error`). ✅
- Standalone unit test of the heuristic: 8/8 (fires on the flood incl. interleaved; never on denylist / distinct / slow-drip / sub-threshold; latches; re-arms).

## Notes / follow-ups

- This is the **error-surfacing** fix. The underlying pipeline attachment mismatch (the actual GPU bug) is separate and engine-side (needs a WASM rebuild).
- Confirmed workaround for the underlying bug: **Settings → Graphic Profile → change it**. Worth offering that as a recovery action in the modal later.
- Longer term, a cleaner engine-side signal (`device.lost` / a latched call from `on_uncaptured_error`) would beat string-sniffing logs.

Stacked on `fix/ui/05-emote-hotkeys` (base). Draft.

### DIFF
--- a/deploy/web/engine/boot.js
+++ b/deploy/web/engine/boot.js
@@ -55,6 +55,22 @@ publish()
   const HANG_TICKS = 8 // ~16s with no corroborating error
   const ERROR_CONFIRM_TICKS = 2 // ~4s after a recorded error
 
+  // "Alive but rendering-dead" detector. The stall watchdog above only catches a FROZEN engine (the
+  // heartbeat stops). A render loop that keeps beating while every frame throws the SAME error — e.g. a
+  // wgpu pipeline/attachment mismatch that blanks the screen and spams the console — sails past it, so
+  // catch it by the flood: the identical engine-core console.error repeating FATAL_REPEATS times within
+  // FATAL_WINDOW_MS ⇒ runtime crash. Scene (UGC) console.error lives in the sandbox WORKER's own
+  // console and never reaches this main-thread patch, so this stream is engine-core only. Signatures are
+  // whitespace-normalized but otherwise EXACT (numbers/urls kept) so distinct errors — different assets,
+  // different coords — never pool into one false flood; only a truly identical per-frame error trips it.
+  const FATAL_REPEATS = 30
+  const FATAL_WINDOW_MS = 4000
+  // Engine-core errors that can repeat identically without being fatal — never trip the flood on these.
+  const BENIGN_REPEATS = ['Message too large']
+  // signature -> { count, since }. Per-signature counters (not a single last-seen key) so an identical
+  // per-frame error still floods even when the loop interleaves it with a second error each frame.
+  const floodCounts = new Map()
+
   const nowMs = () => (window.performance?.now ? performance.now() : Date.now())
   const errMessage = (err) => {
     if (!err) return 'Unknown error'
@@ -80,6 +96,7 @@ publish()
     missedTicks = 0
     recoverTicks = 0
     lastBeatCount = beatCount
+    floodCounts.clear()
     window.__bevyPanic = undefined
   }
   window.__rearmCrashWatchdog = rearm
@@ -90,6 +107,28 @@ publish()
   }
   window.reportEngineError = recordError
 
+  // Collapse whitespace only (keep numbers/urls) so "the same error" groups but distinct ones don't.
+  const signature = (msg) => String(msg).replace(/\s+/g, ' ').trim().slice(0, 300)
+
+  // Flood detector (see FATAL_REPEATS above): an identical engine-core error repeating every frame is a
+  // running-but-dead loop the stall watchdog can't see. Latched via `shown`; rearm() clears the counters.
+  function considerFatalRepeat(msg) {
+    if (shown) return
+    if (BENIGN_REPEATS.some((b) => msg.indexOf(b) !== -1)) return
+    const key = signature(msg)
+    const now = nowMs()
+    let e = floodCounts.get(key)
+    if (!e || now - e.since > FATAL_WINDOW_MS) {
+      e = { count: 0, since: now }
+      floodCounts.set(key, e)
+    }
+    if (++e.count >= FATAL_REPEATS) crash(msg, 'runtime')
+    // Bound memory: drop signatures whose window has lapsed once the map grows.
+    if (floodCounts.size > 64) {
+      for (const [k, v] of floodCounts) if (now - v.since > FATAL_WINDOW_MS) floodCounts.delete(k)
+    }
+  }
+
   // Capture Rust panic text: the wasm panic hook prints "panicked at …" via console.error, while
   // the throw that surfaces is a generic trap — stash the readable message for the host.
   const origConsoleError = console.error.bind(console)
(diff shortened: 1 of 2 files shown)


gonpombo8 (changes requested):
**TL;DR: the detector itself is well-built, but it's wired to the wrong channel — the engine's `error!()` log lines never go through `console.error` on web, so the exact wgpu flood that motivated this PR won't be caught.** The E2E test passed because it spammed `console.error` by hand, which is not the path the real error takes.

## Blocker: engine-core errors arrive via `console.log`, not `console.error`

The premise "Engine-core log lines arrive as a single formatted string (wasm `__wbg_error` → console.error)" holds for the **panic hook** only. Regular `error!()` logs — including `captured wgpu error: {e:?}` from `src/lib.rs:505` — take a different route:

- `src/web.rs:135` passes `custom_layer: |_| None`, so bevy_log's wasm path uses `tracing_wasm::WASMLayer` with default config (Cargo.lock pins tracing-wasm 0.2.1).
- tracing-wasm 0.2.1 only binds `js_name = log` — every level, ERROR included, is emitted as a 4-arg `console.log("%cERROR%c src/lib.rs:505 …captured wgpu error: …", css, css, css)`. It has no `console.error` binding at all.
- The built glue confirms it: `deploy/web/engine/pkg/webgpu_build.js:2985` is the 4-arg `console.log` import; the only `console.error` imports (lines 2423/2429) are the `__wbg_error`/panic-hook path.

So the patched `console.error` will see panics (as before) and the watchdog's own lines, but the per-frame wgpu validation spam sails past it — same silent black screen as in the 2026-07-17 playtest. (Chrome's own native GPU-validation console messages are browser-internal and can't be hooked from JS either.)

Fix within the same host-side / no-WASM-rebuild constraint: also patch `console.log` and feed `considerFatalRepeat` only when the first arg is a string starting with `%cERROR` (tracing-wasm's stable prefix — the `file:line` origin is identical across repeats, so exact-signature matching still works). Everything else about the detector can stay as is.

## Medium: the latch is defeated by the watchdog's auto-rearm — in exactly the target scenario

`boot.js` L126–130 (pre-existing): while `shown` is set, two consecutive ticks with an advancing heartbeat are treated as "transient stall recovered" and call `rearm()`. But the whole premise of the flood case is that **the heartbeat keeps beating**. So ~4s after the flood trips `crash()`, the watchdog auto-rearms, clears `floodCounts`, the still-ongoing flood re-trips within ~0.5s, and the cycle repeats — `__onEngineCrash` fires every ~4.5s for as long as the error persists.

It's not user-visible today because the host handler is idempotent (`react-web/src/features/session/useEngineSession.ts` — `setFatalError(prev => prev ?? …)` keeps the first modal), but "latched, fires once" only holds host-side, and each cycle re-spams `[crash watchdog]` to the console. Consider giving the flood path its own latch that only `__rearmCrashWatchdog` (host dismiss) clears, rather than sharing `shown` with the frames-resumed recovery path — frames resuming is precisely *not* evidence of recovery here.

## Minor

- **300-char signature truncation can pool distinct errors** (`signature()`): wgpu validation messages are verbose with long shared prefixes; 30 distinct-but-same-first-300-chars errors within 4s (e.g. per-asset failures differing only in a trailing label) would false-trip a crash modal on a healthy engine. Cheap hardening: hash the full normalized string, or lengthen the slice.
- Only `arguments[0]` is inspected, so multi-arg calls like `console.error('[Main JS] X worker crashed', e)` (`engine.js:23`) are keyed on the stable first string — a worker crash-loop would trip the flood. Arguably desirable, but worth knowing it's in scope.

## What's good

The detector mechanics are solid: per-signature counters correctly handle interleaved errors, the window-lapse reset gives clean "N within T" semantics, `rearm()` clearing `floodCounts` integrates with the existing dismiss flow, the `[engine error]`/`[crash watchdog]` self-trip guards are correct (both re-entry paths traced), and the scene-sandbox claim checks out — UGC `console.error` stays in the worker's console. The PR description and testing writeup are exemplary; the testing just validated the mechanism rather than the real signal path, which is how the channel mismatch slipped through.

Requesting changes for the `console.log` channel fix (small, and stays within this PR's host-side-only scope); worth folding the latch fix in at the same time since it only shows up once the real flood actually reaches the detector.

pabloes:
* channel fixed: detector listens on `console.log` (tracing-wasm has no `console.error` binding at all)
* matching on `captured wgpu error` instead of the `%cERROR` prefix, which false-positives on healthy sessions (`bevy_asset` 404s, comms `channel closed`)
* flood crash now has its own latch, so the "frames resumed" auto-rearm can't clear it
* backlog *7: boot/init failure leaves an infinite loading bar with no popup, bigger catch than this one

gonpombo8 (approved):

PR #999, opened 2026-07-24
imposters: configurable imposter source url

eordano:
# imposters: configurable imposter source url

## What

`--imposter-source <url>` (native) / `?imposterSource=<url>` (web): the base url of the imposter store the client fetches tiles from. Defaults to the shared store (`https://bevy-imposters.dclregenesislabs.xyz`). Only the base is replaced — the realm-keyed path beneath it (`/imposters/realms/{realm}/{level}/{x},{y}.{crc}.zip`) is the same as the default store's, so a custom store is laid out exactly like the shared one.

```
bevy-explorer --server https://my-realm.example.com --imposter-source https://my-imposters.example.com
```

## Why

The shared store already serves any number of realms (the fetch path is keyed by the current realm's `about_url`, and the crc handles staleness). What it can't do is serve a realm nobody has baked. This lets a custom realm bring its own store — a mirror of the shared layout, or a store baked for that realm — instead of missing on every tile.

## Notes

- Base-url-only (rather than replacing everything up to the realm segment, as first proposed) keeps the store multi-realm and realm-switch-safe and spares the operator the double-encoded realm segment. Discussed in the thread below.
- Startup-only (a `OnceLock` set next to the pulse endpoint override), not an `AppConfig` field: the merged config is written back on any settings/permission change, which would have persisted a one-off flag into the saved config.
- Web follows the `?pulseServer=` pass-through (`EngineHost` → `__bevyBootConfig` → `boot.js` → `engine.js` → `engine_run`). Not behind the untrusted-launch gate: the value is a fetch base for cosmetic imposter zips, no session data involved.
- Both stores keep writing into the same realm-keyed local cache; a tile fetched from a custom store stays cached for that realm+crc until the cache is cleared.

### DIFF
--- a/crates/common/src/structs.rs
+++ b/crates/common/src/structs.rs
@@ -458,6 +458,10 @@ pub struct AppConfig {
     pub inputs: InputMapSerialized,
     pub point_at_marker_visibility: PointAtMarkerVisibility,
     pub camera_smoothing: CameraSmoothing,
+    // where imposter tiles are fetched from: url of the store directory whose
+    // {level}/{x},{y}.{crc}.zip tiles cover the current realm. None = the
+    // default store, keyed by the current realm
+    pub imposter_source: Option<String>,
     // field-level default (0) so configs saved before this field existed read as outdated,
     // rather than taking the current generation from the container-level default
     #[serde(default)]
@@ -499,6 +503,7 @@ impl Default for AppConfig {
             inputs: Default::default(),
             point_at_marker_visibility: Default::default(),
             camera_smoothing: Default::default(),
+            imposter_source: None,
             settings_generation: SETTINGS_GENERATION,
         }
     }
--- a/crates/common/src/structs.rs
+++ b/crates/common/src/structs.rs
@@ -513,6 +513,10 @@ pub struct AppConfig {
     pub inputs: InputMapSerialized,
     pub point_at_marker_visibility: PointAtMarkerVisibility,
     pub camera_smoothing: CameraSmoothing,
+    // where imposter tiles are fetched from: url of the store directory whose
+    // {level}/{x},{y}.{crc}.zip tiles cover the current realm. None = the
+    // default store, keyed by the current realm
+    pub imposter_source: Option<String>,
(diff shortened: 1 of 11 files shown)


robtfm:
Thanks for this! The mechanism is useful but I think both the framing and the URL construction need a rework.

**The current store already serves multiple realms.** The fetch path is keyed by the current realm's `about_url` (`{base}/imposters/realms/{encoded about_url}/{level}/{x},{y}.{crc}.zip`), so the store can hold tiles for any number of realms, and the crc (derived from the scene set the baker observed vs the scenes the client sees) handles staleness — mismatch falls back to an empty imposter by design. A realm doesn't 404 because the store "was baked for" a different realm; it 404s because nobody has baked that realm yet. So the real value here is letting a custom realm bring its own imposter store when the shared one hasn't baked it.

**Suggestion: override only the base URL, keep the path construction.** As written, `imposter_source` replaces everything up to the realm segment, which has a few downsides:

- the request no longer carries the realm key, so a custom store can only ever serve a single realm;
- `imposter_source` is static config but the realm can change at runtime — after a realm switch the client keeps fetching from the old realm's directory (empty imposters at best, wrong imposters if crcs happen to collide);
- the operator has to hand-construct the double-encoded realm segment (`%253A...`), and getting it wrong just 404s silently.

If instead the config replaces only `DEFAULT_IMPOSTER_URL_BASE` and the existing realm-keyed path construction stays as-is, the custom store behaves exactly like the default one (multi-realm, realm-switch-safe, no encoding contract), the diff shrinks to a couple of lines, and the invocation becomes just `--imposter-source https://my-imposters.example.com`.

If I don't hear back in a week or two I'll assume replacing the base url is acceptable and update the PR accordingly.

robtfm:
Updated as discussed: `--imposter-source` now replaces only the store base url and keeps the realm-keyed path (your commit is kept underneath, rebased onto main). Two additions on top: it's a startup-only override rather than an `AppConfig` field (the merged config gets written back on settings changes, which would have persisted the flag), and web parity via `?imposterSource=`. PR description rewritten to match.

PR #1002, opened 2026-07-24
render: gate fixed GPU costs (RTT cameras, sky compute, UI-only cameras)

eordano:
# render: gate fixed GPU costs (RTT cameras, sky compute, per-layer shadows)

## What

Cuts several GPU costs that were paid every frame regardless of whether anything on screen needed them:

- **Texture cameras** (`PbTextureCamera`): a scene could keep any number of RTT cameras active, each one a full render pass with its own whole-world visibility pass. Cameras in the current scenes are now capped at 4 active per scene, preferring the most recently requested (the existing `TODO: limit / cycle` in `update_texture_cameras`).
- **Photo booth cameras**: booth cameras spawned with `is_active: true` and stayed active for the life of the booth, so a backpack/emotes/profile avatar kept re-rendering even when no UI was showing it. They are now active only while some `BoothImage` UI node is visible with a non-zero layout size. Snapshots are unaffected (they use their own transient cameras).
- **World-UI RTT cameras** (text shapes, scene UI canvases): these `Camera2d`s only render UI, but with default `RenderLayers` each one still ran a whole-world visibility pass over layer-0 entities. They now spawn with `RenderLayers::none()`.
- **Sky compute**: `apply_global_light` took `AtmosphereMut<NishitaCloud>` every frame, and merely constructing that param marks the atmosphere resource changed — re-dispatching the `bevy_atmosphere` skybox compute every frame even with an unchanged sky. The atmosphere update is split into its own system that runs every 8 frames, with epsilon jump-detection on `SceneGlobalLight` and a 64-frame full-rate burst after a real jump.
- **Per-layer suns**: every extra per-render-layer directional light spawned with the same shadow settings as the primary sun, each adding a full shadow cascade pass for a light that only exists to illuminate an RTT layer. Extra suns are now shadowless, with a self-heal guard in case something (e.g. the `/shadows` console command) re-enables them.
- **Shadow map size**: the directional shadow map is sized from graphics settings instead of a hard-coded 4096: on web it drops to 2048 unless shadow quality is set to High, and resizes when the setting changes.

## Why

These are fixed per-frame costs: they scale with what content declares, not with what is visible, and they persist when the world is completely still. On WebGPU they were among the largest contributors to steady-state frame time.

This change shipped as part of an engine performance pass measured with paired interleaved A/B runs (headed Chromium/WebGPU on a discrete NVIDIA GPU, Mann-Whitney gated): on a 17-scene benchmark realm orbit, cpu p50 -2.7% (p=.028) and wall p50 -6.9% (7/7 rounds); with 32 synthetic avatars added, cpu p50 -9.2% (p=.029) and wall p50 -49.7% (62.6 ms -> 31.5 ms, p=.025).

## How

Three commits — the first two are independent of each other; the third only names the tuning constants the second introduced (`SKY_JUMP_*` thresholds, shadow map sizes):

**`gate rtt cameras on content and consumer activity`**

- `crates/texture_camera/src/lib.rs`: new `TextureCamRequest(u32)` component records the frame at which each camera was last (re)created. `update_texture_cameras` collects existing cameras per scene root, sorts by request recency, and `set_if_neq`s `is_active` so only the top `MAX_ACTIVE_TEXTURE_CAMERAS_PER_SCENE = 4` in an active scene render.
- `crates/avatar/src/avatar_texture.rs`: new `update_booth_camera_activity` system queries `(&BoothInstance, &InheritedVisibility, &ComputedNode)` on `BoothImage` nodes and drives each booth camera's `is_active` from consumer visibility. A `Local` set of managed cameras deactivates cameras whose last consumer disappeared.
- `crates/world_ui/src/lib.rs`: `spawn_world_ui_view` adds `RenderLayers::none()` to the camera (covers both text-shape views and the scene-UI canvas camera).

**`throttle sky compute updates and trim fixed shadow costs`**

- `crates/visuals/src/lib.rs`: the atmosphere writes move from `apply_global_light` into a new `update_atmosphere` system gated by a `sky_needs_update` run condition (pure `Local` state: frame counter, jump detection, burst countdown). A new `BlendedGlobalLight` resource carries the transition-blended light from `apply_global_light` to `update_atmosphere`; cloud animation uses real elapsed time between runs so cloud speed is unchanged.
- Same file: extra per-layer suns spawn with `shadows_enabled: false`, plus the self-heal loop; `DirectionalLightShadowMap` is initialized from `directional_shadow_map_size(&config.graphics)` and updated by a `resource_changed::<AppConfig>`-gated system.

## Default behavior

Visual output is unchanged at steady state. The sky is fully redrawn on every dispatch, so a throttled update is a complete (never partial) skybox; between updates the last skybox persists, and the dithered sky hides the 8-frame step during slow drift (day/night cycle). A real lighting jump (scene light override, teleport) triggers the full-rate burst, so the sky snaps within a frame of the jump being detected. Scenes with more than 4 texture cameras see the least recently requested ones freeze until re-requested (previously all rendered every frame); avatar booth output is pixel-identical whenever it is on screen.

## Testing

- `cargo check --workspace` clean on this branch.
- The same gating was soak-tested in a headed client on the tree this is ported from: texture-camera scenes render and video textures update; booth images render and drag-rotate while open, and their cameras deactivate when no UI shows them; text shapes and scene UI canvases render unchanged; the day/night sky drifts smoothly and snaps on scene light overrides; shadows are unchanged for the primary sun.
- The perf numbers above come from paired interleaved A/B runs of the containing performance pass on a 17-scene benchmark realm.

## Evidence

Independently verified instrumented A/B pass on this branch alone (not the fork series) against the base commit, audited against the raw artifacts with every count recomputed. (Gathered before the final constant-naming commit, which changes no values — `cargo check` re-run clean on the final tip.)

**Method.** Both sides carry a byte-identical instrumentation overlay counting, per frame in the main world, (a) atmosphere compute dispatches — via the atmosphere resource's change detection, which is exactly the mechanism that re-dispatches the `bevy_atmosphere` skybox compute — and (b) active render-to-texture cameras. These are deterministic main-world proxies for the GPU passes, identical on both sides, not GPU-side pass instrumentation. Two run families on a local mirror of the scene-explorer-tests realm, headed on a discrete NVIDIA GPU, checksum-verified binaries, fresh process per run: interleaved idle-window runs (60 s idle + 30 s moving, per-window counter deltas) and 6 interleaved wall rounds (32 synthetic avatars, 900 measured / 300 warmup frames, exact two-sided Mann-Whitney on per-round medians).

**Results — sky compute gating (confirmed, exact counts).**
- Base dispatches the atmosphere compute every frame, exactly: dispatches == frames in all 4 idle/move windows and all 6 wall rounds.
- The branch dispatches exactly 1 per 8 frames when lighting is static: observed 1/7.96–1/8.02 in all 4 windows, idle and moving alike.
- The full-rate burst fires exactly once per run: in every wall round the branch's dispatch total exceeds frames/8 by 56.1–56.9, and one 64-frame burst contributes exactly 64 − 64/8 = 56 — the boot-teleport lighting jump, followed by steady 1/8.

**Results — RTT camera gating (code-verified only).** The benchmark realm instantiates no photo booth and no scene texture cameras, so booth deactivation and the 4-per-scene cap could not fire at runtime. The counter shows exactly one active RTT camera per frame on both sides — the world-UI view camera, which this branch intentionally keeps active with `RenderLayers::none()` (its layers are emptied rather than the camera deactivated). The gating itself is verified in the diff: cap = 4 ordered by request recency, `RenderLayers::none()` on the world-UI camera bundle, booth `is_active` toggled by visible UI consumers.

**Wall/cpu (no claim made).** wall p50 +3.9% (p = 0.394), cpu p50 +16.9% (p = 0.310) — not significant in either direction. Two branch-side rounds coincided with unrelated workloads on the shared verification GPU, and the saved work (7/8 of one fullscreen compute dispatch) is below that rig's contention-shaped noise floor. The deterministic counters above are the verified signal.

**Regression gate.** `cargo test --workspace` in a private target dir: 80 suites ok, exactly one failure (`scene_runner test::cyclic_recovery`), identical to the pre-existing base baseline. No new failures.

**Boot smoke.** Plain branch binary, default config, 90 s: scenes start and tick (3,359 frames at cutoff), non-black screenshot, no fatal panic — only a known environmental audio-thread panic pair also present in base runs.

**Gaps (stated plainly).**
- The RTT gating half (booth cameras, per-scene cap) has zero runtime evidence in this pass — the benchmark realm has no RTT content. It is code-verified only.
- No scene in the benchmark realm changes lighting mid-run, so the burst was exercised only by the boot-teleport jump, not an interactive mid-run lighting change.
- The per-layer-sun shadow and shadow-map-size changes were not separately instrumented in this pass.

### DIFF
--- a/crates/avatar/src/avatar_texture.rs
+++ b/crates/avatar/src/avatar_texture.rs
@@ -6,6 +6,7 @@ use bevy::{
     diagnostic::FrameCount,
     ecs::system::SystemParam,
     math::FloatOrd,
+    platform::collections::{HashMap, HashSet},
     prelude::*,
     render::{
         camera::RenderTarget,
@@ -39,7 +40,15 @@ impl Plugin for AvatarTexturePlugin {
     fn build(&self, app: &mut App) {
         app.init_resource::<LiveBooths>();
         app.add_systems(Startup, setup.in_set(SetupSets::Main));
-        app.add_systems(Update, (update_booth_image, snapshot, clean_booths));
+        app.add_systems(
+            Update,
+            (
+                update_booth_image,
+                update_booth_camera_activity,
+                snapshot,
+                clean_booths,
+            ),
+        );
     }
 }
 
@@ -305,6 +314,37 @@ fn update_booth_image(
                 height: (node_size.y as u32).max(16),
                 ..Default::default()
             };
+        }
+    }
+}
+
+// booth cameras render only while some ui node is actually displaying their output
+// (snapshots use their own transient cameras, so they are unaffected)
+fn update_booth_camera_activity(
+    consumers: Query<(&BoothInstance, &InheritedVisibility, &ComputedNode), With<BoothImage>>,
+    mut cameras: Query<&mut Camera>,
+    mut managed: Local<HashSet<Entity>>,
+    mut desired: Local<HashMap<Entity, bool>>,
+) {
+    desired.clear();
+    for camera in managed.iter() {
+        desired.insert(*camera, false);
+    }
+    for (instance, visibility, node) in consumers.iter() {
+        let shown = visibility.get() && node.size().cmpgt(Vec2::ZERO).all();
+        let entry = desired.entry(instance.camera).or_insert(false);
+        if shown {
+            *entry = true;
+        }
+    }
+    managed.clear();
+    for (camera_ent, active) in desired.iter() {
+        let Ok(mut camera) = cameras.get_mut(*camera_ent) else {
+            continue;
+        };
+        managed.insert(*camera_ent);
+        if camera.is_active != *active {
+            camera.is_active = *active;
         }
     }
 }
--- a/crates/avatar/src/avatar_texture.rs
+++ b/crates/avatar/src/avatar_texture.rs
@@ -345,7 +345,9 @@ fn update_booth_camera_activity(
             continue;
         };
         managed.insert(*camera_ent);
-        camera.map_unchanged(|c| &mut c.is_active).set_if_neq(*active);
+        camera
+            .map_unchanged(|c| &mut c.is_active)
+            .set_if_neq(*active);
     }
 }
 
--- a/crates/avatar/src/avatar_texture.rs
+++ b/crates/avatar/src/avatar_texture.rs
@@ -322,6 +322,8 @@ fn update_booth_image(
 // (snapshots use their own transient cameras, so they are unaffected)
 fn update_booth_camera_activity(
     consumers: Query<(&BoothInstance, &InheritedVisibility, &ComputedNode), With<BoothImage>>,
+    // shares `&mut Camera` with `update_texture_cameras`; the two touch disjoint camera
+    // entities so bevy just orders them (no marker exists to make the access disjoint)
     mut cameras: Query<&mut Camera>,
     mut managed: Local<HashSet<Entity>>,
     mut desired: Local<HashMap<Entity, bool>>,
@@ -339,13 +341,11 @@ fn update_booth_camera_activity(
     }
     managed.clear();
     for (camera_ent, active) in desired.iter() {
-        let Ok(mut camera) = cameras.get_mut(*camera_ent) else {
+        let Ok(camera) = cameras.get_mut(*camera_ent) else {
             continue;
         };
         managed.insert(*camera_ent);
-        if camera.is_active != *active {
-            camera.is_active = *active;
-        }
+        camera.map_unchanged(|c| &mut c.is_active).set_if_neq(*active);
     }
 }
 
--- a/crates/texture_camera/src/lib.rs
+++ b/crates/texture_camera/src/lib.rs
@@ -3,6 +3,7 @@ use std::f32::consts::FRAC_PI_4;
 use bevy::{
     app::{HierarchyPropagatePlugin, Propagate, PropagateSet, PropagateStop},
     asset::RenderAssetTransferPriority,
+    diagnostic::FrameCount,
     platform::collections::{HashMap, HashSet},
     prelude::*,
     render::{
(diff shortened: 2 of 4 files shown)


robtfm:
Thanks for the pass — one of these is a clear win, but after review we're only taking that piece forward, so closing this PR. Per-change notes:

**World-UI `RenderLayers::none()`** — good and useful; extracted into its own PR: #1075.

**Texture-camera cap** — a hard cap that permanently freezes the least recently *requested* cameras isn't the behaviour we want: a scene legitimately using more than 4 live cameras (e.g. surveillance-monitor setups) silently loses the older ones, since the request stamp only refreshes when the scene rewrites the component. The direction we want is the "cycle" half of the old TODO: round-robin the over-limit cameras so nothing freezes permanently. Follow-up issue: #1073.

**Booth camera gating** — booths are only reachable via the legacy/builtin UI, so there's no real-world win here to justify the extra machinery (and its implicit contract that every consumer is a `BoothImage`).

**Sky compute throttle** — the premise doesn't hold: the sky is *not* fully redrawn on every dispatch. `nishita_cloud.wgsl` updates 1 texel per 8×8 block per dispatch (`let tick = nishita.tick & 63;` walking a dither pattern), so the per-frame cost is already amortized 64× and a full sky refresh takes 64 dispatches. Throttling to 1/8 stretches full refresh to ~512 frames during drift, and the 1s full-rate burst (~60 frames) doesn't complete even one refresh after a lighting jump — so "the sky snaps within a frame" doesn't hold either. The saving is 7/8 of an already-1/64-sized compute, consistent with the branch-alone verification finding no significant wall/cpu difference. Not a good trade against slower sky convergence.

**Per-layer sun shadows / shadow map size** — reducing the web shadow map to 2048 on top of the existing single-cascade low mode pushes quality (already very low at the default shadow distance) below acceptable without also shrinking distance further; not a tradeoff we want to make. On the extra per-layer suns: hard-disabling them (with a self-heal that fights anything re-enabling them) isn't the right mechanism — they should participate in the existing `shadow_caster_count` budget that `manage_shadow_casters` applies to scene lights (which currently only covers point/spot lights, not directional). Follow-up issue: #1074. Holding off on changing sun shadow behaviour until that's done properly.

PR #1007, opened 2026-07-24
scene runner: watchdog for non-responding scene workers

eordano:
# scene runner: watchdog for non-responding scene workers

## What

Enforce the existing `SCENE_NOT_RESPONDING_TIMEOUT` on the renderer side: a scene whose tick has been in flight longer than the timeout is marked `broken`, its `jobs_in_flight` slot is freed, and any reply that arrives after the mark is discarded so it cannot un-break the scene.

## Why

The timeout constant already exists and drives UI (the "not responding" countdown in the out-of-world overlay and the teleport checks), but nothing enforces it. If a scene worker wedges — an infinite loop in `onUpdate`, a runtime hang, a deadlocked worker thread — its `in_flight` flag stays set forever and its entity stays in `jobs_in_flight`. With the default `scene_threads = 4`, one wedged worker permanently consumes a quarter of scene tick throughput, and every other scene is silently throttled for the rest of the session. Two wedged workers halve it, and so on. The user sees nothing except the whole world gradually ticking slower.

Dropping late replies matters for the same reason: if a wedged worker eventually wakes up and flushes a stale `SceneResponse::Ok`, processing it would clear `in_flight` and apply a CRDT state the renderer has long moved past. The scene must stay broken once the watchdog has fired.

## How

- `update_scene_priority` (already the system that maintains `jobs_in_flight`) gets a pre-pass: any scene with `in_flight` set, not already broken, and `elapsed - last_sent > SCENE_NOT_RESPONDING_TIMEOUT` is marked broken with a `warn!`, its `in_flight` cleared, and its entity removed from `jobs_in_flight`. Scenes with an attached inspector are exempt, since a paused inspector session legitimately holds a tick open indefinitely.
- `receive_scene_updates` wraps the `SceneResponse::Ok` processing in a `context.broken` check: late replies from a watchdogged scene are logged at debug level and dropped — no tick increment, no CRDT application, no RPC dispatch. The job-completion bookkeeping after the match still runs, so the slot stays consistent either way.
- The comment on `SCENE_NOT_RESPONDING_TIMEOUT` is updated to describe the enforcement.

No new constants, no config, no protocol changes.

## Default behavior

On by default; there is nothing to configure. Healthy scenes are unaffected — the pre-pass only touches scenes that have already blown a 10 s tick deadline, which is far beyond anything a live scene does. Error handling for scenes that report failure (`SceneResponse::Error`) is unchanged; this covers the workers that never report anything at all. As a side effect, a scene broken via `SceneResponse::Error` that later flushes a stale `Ok` no longer has that reply applied.

## Testing

- `cargo check --workspace` passes.
- Manual: wedge a scene (infinite loop in `onUpdate`) and walk the world. Before this patch the wedged scene holds a `scene_threads` slot forever and neighboring scenes tick slower; with it, the scene is marked broken after 10 s (warn log), the slot is freed, and neighbors are unaffected. If the wedged worker later flushes a reply, it is discarded (debug log) and the scene stays broken.
- This logic has been running in a downstream fork of the engine under heavy automated scene load without regressions.

## Evidence

Independently verified A/B pass on this branch alone against the base commit, re-audited against the raw logs with all counts recomputed. (An earlier pair was invalid — both sides were identity-starved so the wedge scenes never spawned; it was diagnosed, archived, and fully redone.)

**Method.** A local mirror of the `scene-explorer-tests` realm (19 scenes) with two scenes replaced by wedged variants of their published `game.js`:

- WEDGE-INF = **NFT Shape Unit Test** at parcel `54,-68`: enters an infinite loop at its second tick, so its worker never answers again;
- WEDGE-SLOW = **Avatar Attach Unit Test** at parcel `52,-68`: busy-waits 25 s once at its second tick — a reply well past the 10 s timeout, to test that late replies do not un-break.

The 16 remaining healthy scene-explorer-tests scenes get a tail-appended tick logger (every 50th `onUpdate`). Both sides spawn at `52,-60` with `--distance 300 --threads 2` (2 scene-thread slots for 18 scenes), 150 s wall each, identical args and environment, deterministic guest identity on both sides, content cache purged before each run. The two binaries differ by exactly this branch's single commit (verified by patch identity). Signals are deterministic log counters; no perf claim is made.

**Results (exact counts, one 150 s run per side).**

| signal | base | branch |
|---|---|---|
| "marking broken" warns | 0 | 2 — both wedge scenes, both at "has not responded for 10s" |
| late-reply discards | 0 | 1 (WEDGE-SLOW's ~25 s-late reply) |
| WEDGE-SLOW after the late reply | reply accepted, scene keeps running | reply discarded, scene stays broken |
| healthy scenes (of 16) that ever ticked | 11 | 16 |
| scheduler skips on `broken` | 0 | 10,133 (wedges priced out of dispatch instead of holding slots) |

Starvation relief on the branch, by parcel (max tick counter, base → branch): `52,-60` 50 → 4,450 (**89x**), `52,-58` 100 → 3,200, `54,-60` 50 → 2,900, `54,-58` 100 → 2,000. On base, 5 of the 16 healthy scenes (`52,-64`, `52,-66`, `54,-62`, `54,-64`, `54,-66`) never ran a single tick in 150 s, and `52,-62` waited 146.9 s for its first tick vs 35.6 s on the branch.

Claim mapping: marked broken at the existing 10 s timeout — confirmed (both warns at 10 s); slot freed — confirmed (after the marks both slots serve healthy scenes and all 16 tick); base permanently loses a slot per wedged worker — confirmed (WEDGE-INF holds `in_flight` to end of run); late replies discarded without un-breaking — confirmed.

**Honest asymmetry.** The run driver began its camera orbit only on the branch side (on base the wedged realm never reached "ready", so the player stayed at spawn). Movement shuffles distance-based tick shares, but it cannot produce the sign-flips above (0 vs 2 marks, 0 vs 1 discard, accepted vs discarded late reply, 11 vs 16 scenes ticking) — those are decided in the first ~40 s while both sides are still at spawn.

**Regression gate.** `cargo test --workspace --no-fail-fast`: exactly one failing target (`scene_runner test::cyclic_recovery`), identical to the pre-existing base baseline. No new failures.

**Healthy-realm smoke.** Branch binary against the unmodified 19-scene scene-explorer-tests realm: 19/19 scenes started, 0 marked broken, ticks advancing on every scene (per-scene tick counts 36..1,802 at cutoff), 900 measured frames at wall p50 25.3 ms — the watchdog changes nothing on a healthy realm.

**Gaps (stated plainly).** The `--threads 2` identity across sides is enforced by the shared run script rather than echoed in the logs; base never exceeding 2 in-flight jobs and the serial first-tick staircase corroborate it. Two identical environmental audio-thread panics occur on each side (unrelated to scene scheduling).

### DIFF
--- a/crates/scene_runner/src/lib.rs
+++ b/crates/scene_runner/src/lib.rs
@@ -440,6 +440,29 @@ fn update_scene_priority(
         .map(|(e, gt)| (containing_scene.get(e), gt.translation()))
         .unwrap_or_default();
 
+    // mark scenes that have been in-flight past the timeout as broken and free their
+    // slot, so a wedged scene worker can't hold a scene thread forever
+    let elapsed = time.elapsed_secs();
+    for (ent, _, mut context, _) in scenes.iter_mut() {
+        if context.in_flight
+            && !context.broken
+            && !context.inspected
+            && elapsed - context.last_sent
+                > renderer_context::SCENE_NOT_RESPONDING_TIMEOUT.as_secs_f32()
+        {
+            warn!(
+                "scene {} ({} @ {}) has not responded for {:.0}s, marking broken",
+                context.hash,
+                context.title,
+                context.base,
+                elapsed - context.last_sent
+            );
+            context.broken = true;
+            context.in_flight = false;
+            updates.jobs_in_flight.remove(&ent);
+        }
+    }
+
     // check all in-flight scenes still exist
     let mut missing_in_flight = updates.jobs_in_flight.clone();
 
@@ -1092,47 +1115,57 @@ fn receive_scene_updates(
                         rpc_calls.len(),
                     );
                     if let Ok(mut context) = scenes.get_mut(*root) {
-                        context.tick_number = context.tick_number.wrapping_add(1);
-                        if context
-                            .refreeze_at_tick
-                            .is_some_and(|t| context.tick_number >= t)
-                        {
-                            context.blocked.insert(renderer_context::FROZEN_BLOCK);
-                            context.refreeze_at_tick = None;
-                        }
-                        context.last_update_dt = runtime.0 - context.total_runtime;
-                        context.total_runtime = runtime.0;
-                        context.last_update_frame = frame.0;
-                        context.in_flight = false;
-                        // extend (not assign) so renderer-side births (the inspector's /new_entity
-                        // adds the freshly-allocated ids straight to `nascent`) aren't clobbered by
-                        // the scene's reported births. `nascent` is drained every lifecycle pass, so
-                        // this matches assignment for scene-reported births.
-                        context.nascent.extend(census.born);
-                        // Merge externally-queued deaths (e.g. /delete_entity) with
-                        // scene-reported deaths, draining the old set so entries are
-                        // only processed once.
-                        let mut died = census.died;
-                        died.extend(std::mem::take(&mut context.death_row));
-                        context.death_row = died;
-                        for message in messages.into_iter() {
-                            context.log(message);
-                        }
-                        // Sync scene timestamps into crdt_store so renderer writes (e.g.
-                        // /set_component) use a base timestamp that wins over the scene's current.
-                        // Must happen before updates_to_entity drains `crdt`.
-                        context.crdt_store.sync_lww_timestamps_from(&crdt);
+                        if context.broken {
+                            // drop late replies from a scene already marked broken so they
+                            // can't un-break it; still fall through to free the job slot
+                            debug!("[{scene_id:?}] discarding update for broken scene");
+                        } else {
+                            context.tick_number = context.tick_number.wrapping_add(1);
+                            if context
+                                .refreeze_at_tick
+                                .is_some_and(|t| context.tick_number >= t)
+                            {
+                                context.blocked.insert(renderer_context::FROZEN_BLOCK);
+                                context.refreeze_at_tick = None;
+                            }
+                            context.last_update_dt = runtime.0 - context.total_runtime;
+                            context.total_runtime = runtime.0;
+                            context.last_update_frame = frame.0;
+                            context.in_flight = false;
+                            // extend (not assign) so renderer-side births (the inspector's /new_entity
+                            // adds the freshly-allocated ids straight to `nascent`) aren't clobbered by
+                            // the scene's reported births. `nascent` is drained every lifecycle pass, so
+                            // this matches assignment for scene-reported births.
+                            context.nascent.extend(census.born);
+                            // Merge externally-queued deaths (e.g. /delete_entity) with
+                            // scene-reported deaths, draining the old set so entries are
+                            // only processed once.
+                            let mut died = census.died;
+                            died.extend(std::mem::take(&mut context.death_row));
+                            context.death_row = died;
+                            for message in messages.into_iter() {
+                                context.log(message);
+                            }
+                            // Sync scene timestamps into crdt_store so renderer writes (e.g.
+                            // /set_component) use a base timestamp that wins over the scene's current.
+                            // Must happen before updates_to_entity drains `crdt`.
+                            context.crdt_store.sync_lww_timestamps_from(&crdt);
 
-                        let mut commands = commands.entity(*root);
-                        for (component_id, interface) in crdt_interfaces.0.iter() {
-                            interface.updates_to_entity(*component_id, &mut crdt, &mut commands);
-                        }
-                        // dcl_assert!(
-                        //     updates.jobs_in_flight.contains(root) || context.tick_number <= 2
-                        // );
+                            let mut commands = commands.entity(*root);
+                            for (component_id, interface) in crdt_interfaces.0.iter() {
+                                interface.updates_to_entity(
+                                    *component_id,
+                                    &mut crdt,
+                                    &mut commands,
+                                );
+                            }
+                            // dcl_assert!(
+                            //     updates.jobs_in_flight.contains(root) || context.tick_number <= 2
+                            // );
 
-                        for rpc_call in rpc_calls {
-                            rpc_call_events.write(rpc_call);
+                            for rpc_call in rpc_calls {
+                                rpc_call_events.write(rpc_call);
+                            }
                         }
                     } else {
                         debug!(
--- a/crates/scene_runner/src/lib.rs
+++ b/crates/scene_runner/src/lib.rs
@@ -431,6 +431,7 @@ fn update_scene_priority(
     mut updates: ResMut<SceneUpdates>,
     time: Res<Time>,
     containing_scene: ContainingScene,
+    mut commands: Commands,
 ) {
     updates.eligible_jobs = 0;
     updates.frames += 1;
@@ -460,6 +461,9 @@ fn update_scene_priority(
             context.broken = true;
             context.in_flight = false;
             updates.jobs_in_flight.remove(&ent);
+            // dropping the thread handle closes the scene's renderer channel, which
+            // signals the scene host to terminate the worker's isolate
+            commands.entity(ent).try_remove::<SceneThreadHandle>();
         }
     }
 
@@ -793,6 +797,7 @@ fn send_scene_updates(
     preview_mode: Res<PreviewMode>,
     mut buf: Local<Vec<u8>>,
     mut realm_info_cache: Local<RealmInfoCache>,
+    mut commands: Commands,
 ) {
     let updates = &mut *updates;
     let buf = &mut *buf;
@@ -1027,7 +1032,7 @@ fn send_scene_updates(
             context.base
         );
         context.broken = true;
-        // TODO: clean up
+        commands.entity(ent).try_remove::<SceneThreadHandle>();
     } else {
         context.in_flight = true;
         context.last_sent = time.elapsed_secs();
@@ -1091,6 +1096,7 @@ fn receive_scene_updates(
                         if let Ok(mut context) = scenes.get_mut(*root) {
                             context.broken = true;
                             context.in_flight = false;
+                            commands.entity(*root).try_remove::<SceneThreadHandle>();
                             let timestamp = context.total_runtime as f64 + 1.0;
                             error!("[{scene_id:?} @ {}] error: {message}", context.tick_number);
                             context.log(SceneLogMessage {
--- a/crates/scene_runner/src/lib.rs
+++ b/crates/scene_runner/src/lib.rs
@@ -56,7 +56,7 @@ use self::{
     initialize_scene::{
         LiveScenes, PointerResult, SceneLifecyclePlugin, SceneLoading, ScenePointers, PARCEL_SIZE,
     },
-    renderer_context::RendererSceneContext,
+    renderer_context::{RendererSceneContext, SceneState},
     update_scene::SceneInputPlugin,
     update_world::{CrdtExtractors, SceneOutputPlugin},
 };
@@ -111,7 +111,7 @@ impl SceneUpdates {
     }
 }
 
-#[derive(Component)]
+#[derive(Debug)]
 pub struct SceneThreadHandle {
     pub sender: tokio::sync::mpsc::UnboundedSender<RendererResponse>,
 }
@@ -431,7 +431,6 @@ fn update_scene_priority(
     mut updates: ResMut<SceneUpdates>,
     time: Res<Time>,
     containing_scene: ContainingScene,
-    mut commands: Commands,
 ) {
     updates.eligible_jobs = 0;
     updates.frames += 1;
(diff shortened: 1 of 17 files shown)


robtfm:
This doesn't terminate the wedged worker, just stops scheduling it — the worker keeps spinning on a core forever, and never notices despawn either, since it never returns to its event loop. Freeing the slot is the right trigger layer, but on its own each wedge still permanently costs a core, and reloading the scene spawns a replacement while the old worker keeps burning. That accumulates badly, especially for server/headless. So I want to add actual termination on top of the `broken` mark before merging.

**native**: `dcl_deno` already stores each isolate's `thread_safe_handle()` in `VM_HANDLES`, but nothing ever calls `terminate_execution()` (and entries are never removed). Plan: a control message over the ipc channel → worker process looks up the handle → `terminate_execution()`. That interrupts even an infinite JS loop; the scene thread unwinds and exits, reclaiming the core and memory. It should also fire on scene despawn, since wedged workers currently outlive their scene. Caveat: it only interrupts JS, not a worker blocked inside a rust op — acceptable, the canonical wedge is scene JS.

**web**: we can't naively `Worker.terminate()` a sandbox worker — it shares the engine's wasm memory, and killing it mid-op while it holds a shared lock (allocator, channel mutex) could corrupt or deadlock the whole engine. Two-tier plan:

- *tier 1 — cooperative*: post a SHUTDOWN message to the worker. The common real-world wedge is the old SDK timer re-arm hang, where the tick loop is stuck at an `await` while timers churn — the event loop is alive, so the message handler runs with an empty JS stack (never mid-rust). It sets a local flag so op wrappers refuse further entry, then self-tears-down cleanly: `__wbindgen_thread_destroy()` + `close()`. Nothing leaks, no forceful kill needed.
- *tier 2 — forceful, only on tier-1 timeout*: no ack means the event loop is dead, i.e. a synchronous spin — and in that state timers/microtasks can't run, so the only route into wasm is the op wrapper in `sandbox_worker.js` (`createJsContext`). Put a two-flag handshake there, JS-side: wrapper does `Atomics.store(in_rust, 1)`, checks the kill flag (if set → park via `Atomics.wait` on a dummy address rather than throwing, since scene code can catch), calls the op, clears the flag. Engine side: set kill flag, wait for `in_rust == 0`, then `terminate()`. Setting the flag in JS *before* the wasm call means there's no window where rust executes unflagged.

The tiers are complementary: async wedges — where promise/timer resumptions re-enter wasm outside the op wrapper and would leak past the handshake — are exactly the cases where tier 1 works; sync spins — where tier 1 can't run — are exactly the cases where the handshake is airtight.

Wiring/cleanup this needs:

- `spawn_and_init_sandbox` currently drops the `Worker` reference and main JS never learns which scene a sandbox serves. Have the worker postMessage its scene id (plus its `tls_base`/`stack_ptr`) after `wasm_init_scene`, keep a sceneId→worker map, and expose a `terminate_sandbox(sceneId)` callback next to `spawn_and_init_sandbox`.
- After a tier-2 terminate, call `__wbindgen_thread_destroy(tls_base, stack_ptr)` from main JS with the saved pointers — otherwise every kill leaks the worker's stack/TLS in shared linear memory, and wasm memory never shrinks, so repeated kills ratchet the footprint.
- Residual race: a scene alternating sync spins with pending timers could have a timer callback enter wasm in the microseconds between the engine reading `in_rust == 0` and the terminate landing. If that ever bites, it's closable by wrapping the worker's task sources (`setTimeout`, `queueMicrotask`, `Promise.prototype.then`, `onmessage`) to park callbacks once the kill flag is set — the sandbox already locks down its globals, so this is mechanical — but I wouldn't gate on it initially.
- If a worker is legitimately blocked *inside* rust (e.g. waiting on a renderer reply that will never come), the engine should close/complete its side of the scene's channels first to flush the worker out of rustland, then proceed with the handshake. If `in_rust` still never clears, log loudly and leave the worker (status quo).

One note in favour of this PR as-is: the late-reply discard is needed regardless of any kill mechanism — there's always a window between the watchdog firing and termination completing where a stale `Ok` could arrive and would otherwise un-break the scene.

robtfm:
I've pushed the termination work described above onto this branch (rebased onto main first). What's added on top of the watchdog:

**Worker termination.** Marking a scene broken — or despawning it — now drops its `SceneThreadHandle`, which closes the scene's renderer channel and flows through the existing `KillScene` ipc message. The scene host then gives the thread a 5s grace period: a healthy scene finishes its in-flight tick and exits through the normal channel-closed path untouched; if the thread is still around after the grace (wedged in js, or stuck awaiting something external), the host force-terminates the isolate. So the wedge case now actually reclaims the core and memory instead of leaving the worker spinning forever, and mid-tick scenes being despawned aren't cut off arbitrarily.

Force-termination pairs v8 `terminate_execution` with a kill flag the scene loop checks: the termination request is consumed once the js stack unwinds, and the loop deliberately doesn't exit on uncaught errors, so without the flag a deterministic wedge would just re-enter its infinite loop on the next tick with nothing left to terminate it.

**`SceneState` enum.** `RendererSceneContext::broken` and the separate `SceneThreadHandle` component are folded into one field: `state: SceneState { Init, Live(handle), Broken }`. `Broken` consumes the handle, so marking a scene broken structurally drops the channel — a future broken-marking site can't forget to signal the kill.

**Heap cap actually ends the scene.** The near-heap-limit callback now also sets the kill flag. Termination alone only unwinds the current js — the heap stays rooted by the scene globals and the loop keeps ticking, so an OOM-capped scene would re-trip the callback and burn cpu forever while holding the full 512MB. With the flag the loop exits and the runtime drop releases the heap.

**`ShuttingDown` → `KillFlag`.** The old op-state marker and the new flag were two identical exits from the scene loop, differing only in who could set them (op state is single-threaded, so the marker couldn't be set by the scene host). Consolidated into one `KillFlag(Arc<AtomicBool>)` held in op state; the ops' cooperative kills and gating semantics are unchanged, and the wasm worker reads the same flag — it's the exit signal the (follow-up) web-side termination will set.

Verified with a deliberately wedged scene (infinite loop in `onUpdate`): watchdog marks it broken at 10s, force-terminate lands exactly at the 5s grace, sidecar cpu drops from a pinned core to idle, engine unaffected. The broken scene stays as an inert tombstone engine-side and never auto-restarts in place, so a deterministic wedge can't kill→respawn→wedge crash-loop; walking away and back (or reload) spawns a fresh worker. Wasm build passes; web-side worker termination is still to come per the plan above.

crates/scene_runner/src/lib.rs
```diff
@@ -440,6 +440,31 @@ fn update_scene_priority(
...
+    let elapsed = time.elapsed_secs();
+    for (ent, _, mut context, _) in scenes.iter_mut() {
+        if context.in_flight
+            && !context.broken()
+            && !context.inspected
+            && elapsed - context.last_sent
+                > renderer_context::SCENE_NOT_RESPONDING_TIMEOUT.as_secs_f32()
+        {
+            warn!(
+                "scene {} ({} @ {}) has not responded for {:.0}s, marking broken",
```
regenesis-claw, line 453: [P1] This can't tell a wedged worker from a scene that is merely slow. A tick stays in flight until the scene calls `crdt_send_to_renderer` at the end of `onUpdate`, so a scene doing `await fetch(...)` in its update against a slow endpoint blows the 10s deadline while being perfectly alive — and unlike the existing UI use of this constant, the consequence here is permanent (`Broken` tombstone, isolate force-terminated 5s later, no in-place restart).

Either gate the kill on a signal that the worker is genuinely unresponsive rather than parked in a rust op, make the kill recoverable with a strike count, or split the kill deadline from the not-responding-UI deadline so the two policies can move independently.

crates/scene_runner/src/lib.rs
```diff
@@ -1092,47 +1120,57 @@ fn receive_scene_updates(
...
-                        // dcl_assert!(
-                        //     updates.jobs_in_flight.contains(root) || context.tick_number <= 2
-                        // );
-
-                        for rpc_call in rpc_calls {
-                            rpc_call_events.write(rpc_call);
+                        if context.broken() {
+                            // drop late replies from a scene already marked broken so they
+                            // can't un-break it; still fall through to free the job slot
+                            debug!("[{scene_id:?}] discarding update for broken scene");
```
regenesis-claw, line 1126: [P2] Discard is right, but `context.in_flight = false` lives in the `else` arm below, so this branch relies on every `Broken`-setting site having already cleared the flag. That holds today (the watchdog and the `Error` arm both clear it) — hoisting the clear above this `if` makes it structural instead of an invariant maintained at three call sites, which is the same reasoning behind folding the handle into `SceneState`.

regenesis-claw (changes requested):
Reviewed the watchdog + termination stack (263/-124 across 16 files). The core design is sound: `SceneState` folding the handle into the state so `Broken` structurally drops the renderer channel is a genuinely better invariant than a bare `broken: bool` + separate component, and it incidentally removes a latent panic (`send_scene_updates` previously did `scenes.get_mut(ent).unwrap()` on a query that required the handle component, so a queued scene without one would have panicked; it now returns early). The late-reply discard is correct and, as noted in the thread, needed regardless of the kill mechanism. Kill-flag bookkeeping checks out: the watchdog clears `in_flight` and frees the `jobs_in_flight` slot before the late reply lands, and the post-match `jobs_in_flight.remove` is idempotent, so no double-free or slot leak.

One thing I'd want resolved before merge, plus some smaller notes.

### [P1] A slow-but-alive scene is now permanently killed, not just flagged

The 10s deadline measures time since the tick was dispatched, and a tick only completes when the scene calls `crdt_send_to_renderer` at the end of `onUpdate`. A scene that `await`s inside `onUpdate` — `fetch`/`signedFetch` against a slow or unreachable endpoint, default timeouts well past 10s — holds its tick open the whole time and is indistinguishable from a wedge to this check. Today that scene is merely slow (the timeout only drives the out-of-world overlay and the teleport checks, both recoverable). After this PR it is marked `Broken`, its isolate is force-terminated 5s later, and per the description it becomes an inert tombstone that never restarts in place — the user has to walk out of range and back. Same for a heavy synchronous first tick on a low-end device.

That is a real behaviour change for scenes that are badly written but functional, and it fails closed with no recovery. Worth deciding explicitly rather than inheriting it from a constant that previously only drove UI. Options, roughly in order of effort:

- distinguish *wedged* from *slow* — the scene thread knows whether it is parked in a rust op (fetch, awaiting the renderer) versus spinning in JS; a coarse heartbeat/op-activity signal from the worker would only kill the genuinely unresponsive case (this is also the distinction the follow-up web plan draws for tier 1 vs tier 2);
- make the watchdog kill recoverable — respawn the scene once instead of tombstoning it, with a strike count so a deterministic wedge can't crash-loop;
- or, minimally, use a separate (larger) constant for *kill* than for the *not responding* UI, so tightening/loosening the UI hint doesn't silently change the kill policy.

Unverified either way from the diff: the A/B evidence covers deliberate wedges only, and the healthy-realm smoke tests scenes that answer within the deadline, so neither run exercises a scene that is slow-but-alive across the 10s boundary.

### [P2] `in_flight` clearing lives inside the non-broken branch

In `receive_scene_updates`, `context.in_flight = false` sits in the `else` arm. Today every path that sets `Broken` also clears `in_flight` (watchdog, `SceneResponse::Error`), so the flag can't leak — but that's an invariant spread across three call sites rather than one, and the whole point of the `SceneState` refactor is to make this structural. Hoisting the clear above the `broken()` check makes the discard branch self-consistent.

### [P2] No automated coverage for either new behaviour

The only test changes here are mechanical `ShuttingDown` -> `KillFlag` renames. The two behaviours worth pinning are cheap to express against `SceneUpdates` + `RendererSceneContext` without a real worker: (a) an in-flight scene past the deadline gets marked broken, has `in_flight` cleared and its entity removed from `jobs_in_flight`, and an inspected scene does not; (b) an `Ok` response for a scene already in `SceneState::Broken` leaves `tick_number` untouched, applies no CRDT and dispatches no RPCs. Both are pure state transitions.

### [P2] Web still has no termination

The watchdog now fires on wasm too, so a wedged web sandbox gets its slot freed but keeps burning — the shared-memory situation from the earlier comment. That's an explicitly stated follow-up, not a blocker, but until it lands the web build gets the scheduling half of the fix only. Worth a tracking issue so it doesn't sit in a PR thread. The wasm `init_scene` passing `Default::default()` as the kill flag (unshared) is fine for now since only `op_continue_running` reads it, but it means the flag is a no-op signal on web until the handshake exists.

### [P2] Minor

- `KillFlag` is read with `op_state.borrow::<KillFlag>()`, which panics if absent, where `has::<ShuttingDown>()` could not. Every live path goes through `init_state`, so this is safe today; the four `create_runtime` test call sites that needed `Default::default()` added are the warning sign that a future runtime-construction path can trip it. A `try_borrow(..).is_some_and(..)` at the read sites, or putting the flag in `create_runtime` rather than `init_state`, would close it.
- `terminate_scene` logs at `warn!` whenever the handle is still present at the grace deadline. For a normal despawn of a scene mid-teardown that's a scary log line for a benign race; consider warning only when the thread is still *executing* (or downgrading to `debug!` for the despawn path).

### Security

No security issues found. No new user-controlled input, no secrets, no network surface. `EngineToScene::KillScene(id)` -> `Entity::from_bits(id)` in the sidecar is the same trust boundary and the same conversion the existing `NewScene`/`SceneUpdate` paths use — parent process only.

### CI

All checks green on 9b8e1eb (Test Suite, Clippy, Rustfmt, Headless Smoke, Build and Deploy Web, toml). Test Scenes skipped.

No `REVIEW.md` at the repo root, so this follows the generic checklist plus the existing conventions in `scene_runner`.

PR #1014, opened 2026-07-27
fix/ui/07-popup-system: PopupHost owns overlay infra (scrim, focus trap, Escape)

pabloes:
## What

- `PopupHost` becomes the single owner of overlay infrastructure for every HUD modal except the crash/realm-error surfaces (see Notes): the scrim (dim + blur, reference-canvas DPI scale), the pop-in entrance animation, the focus trap, and a single capture-phase Escape listener.
- Every popup consumer (`Modal`/`showDialog`/`showConfirm`, `ExitConfirm`, `Passport`, `ProfileCard`, `PermissionDialog`, `CommunityModal`, `WorldVisitModal`, `CommunityCreateModal`) drops its own `createPortal`/scrim/self-scale/animation/Escape-handler and opens through `openPopup` (or a thin wrapper: `openExitConfirm`, `openPermissionDialog`, `openWorldVisit`, `openCommunityModal`, `openCommunityCreateModal`).
- `Modal` is stripped to a pure card — no portal, no scrim, no `scrimless` flag. `ModalShell` still wraps it for header/body/footer layout.
- `WorldVisitModal` picked up a description + cover image from the places API while it was being migrated (not otherwise related, but it was already being touched).
- Drops the now-redundant engine `Cancel` → `closeTopPopup` relay.
- The old `EngineErrorModal` is split into `CrashModal` (full-screen, self-contained, fatal — boot/runtime engine failures) and `RealmErrorModal` (a normal popup-layer dialog for "world not found"), plus a module-level `inputLock.ts` so a crash appearing over an already-open popup can't double-fire Escape (crash dismisses on its own Escape without also closing the popup underneath, and vice versa).
- Fixed `popups.module.css`'s `.dim` (and the equivalent `CrashModal.module.css` `.scrim`) padding: it was `3vh 3vw` inside a box that's already DPI-scaled to the reference canvas, so real viewport units ignored the ancestor transform; switched to fixed reference px, same pattern as `Modal.module.css`'s card cap.
- Added tier-1.5 Playwright visual baselines for the popups migrated in this PR: Passport, PermissionDialog, RealmErrorModal, CommunityModal, CommunityCreateModal, ExitConfirm (ProfileCard and the crash modal already had one).

## Why

The popup system had accreted a portal + scrim + focus trap + Escape handler per component, each slightly different (different animation timings, some `scrimless`, some not, ExitConfirm on a 0.3s linear transition vs. Passport on 0.2s ease). Centralizing all of that in `PopupHost` means every popup gets the same entrance, the same DPI-scaled scrim, and Escape/focus-trap behavior for free, and new popups don't need to reimplement any of it.

`CommunityModal` needed a different shape than the fire-once popups (`ExitConfirm`, `PermissionDialog`, `WorldVisitModal`, `CommunityCreateModal`): its `detail` arrives async and membership can change mid-open (Join/Leave), so a static `openPopup` closure would freeze at the moment it opened. A `CommunityModalContainer` reads `useSession()` instead, matching the existing `ProfileCard`/`Passport` pattern, so its content stays live.

The engine's `Cancel` system action only fires while the engine canvas holds keyboard focus, but every popup opens with the cursor already freed (the engine clears pointer-lock on the trigger click) — so `PopupHost`'s own DOM Escape listener always fires first, and the relay was dead weight.

The crash/realm-error split exists because a crash can legitimately appear while a popup is already open (e.g. mid-session), and the old single `EngineErrorModal` + a bare Escape listener per surface let one Escape close both at once — `inputLock.ts` is a small counted lock both `PopupHost` and the crash surface check before acting.

## Tests

- `tsc --noEmit` clean throughout.
- Full vitest suite: 293/293 passing (includes a regression test for the Escape double-close scenario and the `ExitConfirm` settle-once guard).
- `npm run test:visual`: 25/25 passing, including the newly added baselines.
- Manual verification in `?mock=1` for each migrated popup: passport, profile card, exit confirm, permission dialog, community modal (join/leave/add-friend/tabs, live update without reopening), world-visit (description/image, broken-image fallback), community-create (NAME gate + form).

## Notes

`CrashModal` (the fatal, full-screen surface) is deliberately left out of the `PopupHost` migration — it renders in the `ErrorBoundary` fallback and the embedded/no-engine mode, where `PopupHost` isn't mounted, so it keeps its own scrim/focus-trap/Escape, coordinated with the popup layer only through `inputLock.ts`. `RealmErrorModal` is a normal `PopupHost` popup.

No known gap besides `WorldVisitModal` visual coverage (tracked as backlog item 41 — needs a places-API stub to be deterministic in mock mode).


gonpombo8 (commented):
**TL;DR: good refactor — the close-contract design is sound and the consumers were migrated carefully. One real bug worth fixing (Escape double-close when a crash modal appears over an open popup), plus a few minor comment/CSS inaccuracies. Nothing blocker-level.**

## Medium: `EngineErrorModal` and `PopupHost` don't coordinate — one Escape closes both layers

Both register `keydown` listeners on `document` in the capture phase (`popups.tsx` L117–127, `EngineErrorModal.tsx` L47–56). `stopPropagation()` only stops propagation to *other nodes* — it does not stop other listeners on the *same* node (that's `stopImmediatePropagation`). So when a dismissable runtime crash modal is up **while any popup is open** (entirely plausible — a crash can land mid-passport or mid-permission-prompt), one Escape press runs both handlers: the top popup closes *and* fires its `onClose` contract (e.g. silently denying a permission, or resolving a `showDialog` with null), and the crash modal dismisses too. The user only sees the crash modal (it's at `--z-fatal`), so the popup underneath closes invisibly.

The two focus traps also both stay active in that state (PopupHost's top-layer trap and `EngineErrorModal`'s own `useFocusTrap(scrimRef, true)`), and they'll fight over Tab since each cycles within its own subtree.

Suggested fix that resolves both at once: clear the popup stack when a fatal modal mounts (a `closeAllPopups()` that walks `closeById` so the `onClose` contracts settle — deny-once / stay / resolve-null are all sensible outcomes under a crash). Then `hasOpenPopup()` is false, PopupHost's Escape handler stands down naturally, and only one trap is active.

## Minor

- **Stale comment contradicting the PR itself**: the `closeTopPopup` doc block (`popups.tsx` L46–49) still says it's fired "as a backup while the engine holds keyboard focus, by the engine's 'Cancel' system action relayed through the bridge — see useEngineSession" — but this PR deletes that relay from `useEngineSession.ts`. The session test now correctly documents Cancel as a no-op; the comment should match. (The removal itself checks out: DOM capture on `document` sees keydown even with the canvas focused, and under pointer lock the browser consumes Escape before *either* path saw it, so behavior there is unchanged.)
- **Wrong CSS claim in `CommunityModal.module.css` L1–2**: it says vw/vh inside the scaled scrim "resolve against that canvas, not the raw viewport". Viewport units always resolve against the real viewport regardless of ancestor transforms — only *percentages* resolve against the scaled box. It happens to work today because every card caps with `min(…px, …)`, but the comment will mislead the next person who sizes a popup in `vh`. Related: the old Modal backdrop's `overflow: auto` has no equivalent on `.dim`, and `Modal.module.css`'s card cap (`calc(100dvh - 40px)`) is in pre-transform units — at `--ui-scale > 1` a tall card can visually exceed the screen with no scroll path. Low practical risk with the current px caps, worth a line of awareness.
- **`openExitConfirm`'s Leave path runs "stay" first**: Leave does `close()` → `closeById` fires `onClose` (= `onStay`) → *then* `onLeave()`. It works only because `stay()` (`setConfirming(false)`) is harmless before `leave()`, and it quietly contradicts the comment "any close path other than Leave keeps the user in-world". The settle-once guard pattern used in `openPermissionDialog`/`showDialog` (mark settled before `close()`) would make this robust against future side effects in `stay`.

## What's good

The linchpin — `closeById` removing the node *before* running `onClose` — makes every close path idempotent, and I traced all the re-entrant cases it protects: effect-cleanup close handles firing after a user close (App's permission/exit effects, CommunitiesPage, MapPage's `closeVisitRef`) are all safe no-ops. The settle-once guards in `showDialog` and `openPermissionDialog` are correct. The `wantOpen` identity check in CommunitiesPage correctly survives the A→B community switch where the old effect's cleanup fires after `openId` already holds the new id. `exitGuard.stay/leave` memoization is exactly what the open-from-effect pattern needs. `CommunityModalContainer` reading `useSession()` (with `PopupHost` mounted inside `SessionProvider`) cleanly solves the frozen-snapshot problem, and `--ui-scale` lives on `documentElement` so both the shared and the self-contained scrim resolve it. The new tests cover the right contracts (Escape layering, pass-through when empty, onClose-exactly-once on all three paths, `showConfirm` resolving false on Escape).

pabloes:
> closeAllPopups

As stated in the commits, instead of `closeAllPopups` I went with canceling hotkeys through an addEventListener wrapper `useWindowKeyDown`, which is UX standard and doesn't lose information for reporting.

Separated EngineErrorModal into different popups, having now a `CrashModal`, as some are fatal and usually unrecoverable errors and others are just information errors that don't stop the app, like failed fetch, etc.

Added note to backlog (*8) to fix in other PR, the issue happens when we type `/goto nonexistentworld` and an infinite reconnecting screen appears.

pabloes:
As stated in previous commits, instead of closeAllPopups I went with an addEventListener wrapper useWindowKeyDown that cancels all hotkeys while a crash modal is up, so no visual information is lost for reporting; separately, menu hotkeys now also stand down while any popup is open (standard modal UX — with a passport open, M was opening the map underneath it).

Separated EngineErrorModal into different popups, having now a CrashModal, as some are fatal and usually unrecoverable errors and others are just information errors that don't stop the app, like failed fetch, etc.

While testing that, found and fixed two layering issues: the popup layer sat below the login/destination picker (new --z-popup token, mirrors bevy-ui-scene's PopupStack at MAX_ZINDEX-1), and popup cards were capped in vw/vh, which ignore the --ui-scale transform (now %).

The three minor items are addressed too: closeTopPopup's stale Cancel-relay comment, the vw/vh claim in CommunityModal.module.css, and openExitConfirm now settles exactly one of stay/leave (settle-once guard + test).

Added note to backlog (*8) to fix in other PR: typing /goto nonexistentworld leaves an infinite reconnecting screen.

gonpombo8 (approved):

PR #1015, opened 2026-07-27
fix: editor scene runs main() once then auto-freezes

nicoecheza:


robtfm:
Reworked the gating while reviewing: the auto-freeze is now driven by an explicit editor flag — `--editor` natively, `editor` in the web boot config/URL — instead of preview||loopback. A plain `npm start` / `--preview` dev scene has nothing attached that would ever unfreeze it, so it must keep free-running; `ShowOutOfBounds` keeps the old broad condition so previews retain out-of-bounds rendering.

**Creator-hub follow-up needed**: pass `editor: true` into the engine boot config where the inspector already wires `systemScene` — `window.__bevyBootConfig` in `packages/inspector/scripts/bevy-engine-host/host-boot.js` / `buildEngineUrl` in `src/lib/renderer/bevy/engine-iframe.ts`. Without it the embedded engine won't freeze at tick 3 anymore. The play/stop bridge (`/freeze_scene` / `/unfreeze_scene`) is unchanged.

Also included: frozen-scene exemptions for the engine's startup tick gates (scene-lifecycle still-loading, imposter live-parcels, out-of-world player release — the latter now also waits for gltf colliders), since a scene frozen at tick 3 never passes them on its own. Rebased onto main.

PR #1021, opened 2026-07-28
fix(comms): LiveKit adapter preserves host:port in room address

eordano:
`initiate_room_connection` rebuilds the room URL from the adapter address with
`Uri::host()`, which returns the host without the port. A livekit address on a
non-default port — a self-hosted or custom SFU — then connects to the default
port instead of the one in the address. Use `Uri::authority()`, which preserves
`host:port`.


robtfm:
fixed driveby in another pr 🤷

PR #1025, opened 2026-07-30
ipc: bound frame length and close instead of panicking on a bad frame

eordano:
Both IPC readers take the 64-bit length prefix on trust and pre-allocate it, then unwrap the body read and the decode. A truncated or corrupt frame panics the reader task; on the engine side that task runs in the process holding the authoritative identity, so the engine and every co-tenant scene go with it.

Frames now go through one `read_frame` helper that rejects a length over `DCL_IPC_MAX_FRAME_BYTES` (256MiB default) and turns a short read or an undecodable body into a clean end of the loop. Failure is terminal for the connection because the stream carries no frame boundary to resynchronise on.

`read_frame` is generic over `AsyncRead + Unpin` so it can be driven from a cursor in tests.

7 tests. Reverting each guard to the current code reproduces the original panic: the length cap removed gives a `raw_vec` capacity overflow, the other two panic at the restored `.unwrap()`.

The 256MiB default is not measured against a real `NewScene` bundle — worth a look if your scenes are large.

### DIFF
--- a/crates/dcl_deno_ipc/src/lib.rs
+++ b/crates/dcl_deno_ipc/src/lib.rs
@@ -82,6 +82,40 @@ pub static NEW_SCENE_SENDER: Lazy<
 /// and for tests, which build a fresh runtime per app and must not kill the process.
 pub static EXIT_ON_SIDECAR_LOSS: AtomicBool = AtomicBool::new(false);
 
+pub fn max_ipc_frame_bytes() -> u64 {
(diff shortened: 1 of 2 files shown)


robtfm:
i don't see much benefit here - closing the ipc connection will tear down the whole process anyway. 

i'll look into some limits at the js ingress boundary instead, so we can keep co-tenanted scenes alive.

PR #1030, opened 2026-07-30
headless: refuse bad arguments instead of silently taking defaults

eordano:
Every option is parsed with `.ok()` or `unwrap_or`, so a value that failed to parse and a value that was absent are indistinguishable. `--tick-hz abc` runs at 30, a malformed `--realm` goes to localhost:8000, and a misspelled flag is discarded — `--server-mod` starts a client. There is no `--help`, so asking for one boots a server against the default realm.

Parsing moves to `opt_value_from_str` so absent and malformed are separate outcomes, leftover arguments are rejected, zero is rejected for the two counts that must not be zero, and `-h`/`--help` prints usage.

`parse_args` is split into `parse_args_from(argv) -> ArgOutcome` so parsing is testable without `process::exit`; the exiting wrapper stays at the boundary.

7 tests, mutation-checked. One mutation initially survived: `--tick-hz abc` was still rejected when parse errors were swallowed, but by the leftover-argument check rather than the parse, so the message was wrong and the test could not tell. It now asserts the rejection says `failed to parse`.

Rejecting unknown arguments is a behaviour change — any harness passing a flag this binary previously ignored now exits 2.

### DIFF
--- a/src/bin/headless.rs
+++ b/src/bin/headless.rs
@@ -88,31 +88,103 @@ struct Args {
     storage_delegation: Option<String>,
 }
 
+const USAGE: &str = "\
+decentra-bevy headless — run scenes with no window
+
+USAGE: headless [OPTIONS]
+
+  --realm <URL>               realm to connect to [default: http://localhost:8000]
+  --location <X,Y>            parcel to stand in [default: 0,0]
+  --preview                   preview mode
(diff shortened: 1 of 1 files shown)


robtfm:
closing in favour of 1219

PR #1031, opened 2026-07-30
scene sandbox: cap WebSocket resources per scene

eordano:
Scene WebSockets have no per-scene bounds: a scene can open sockets without limit, stall a handshake indefinitely, buffer unboundedly on send, and receive messages of any size. On a server hosting many scenes in one process each of those is shared with every co-tenant.

Caps the socket count per scene (32, with in-flight handshakes reserved so a connect burst cannot race past the cap), the handshake wait (15s), the outbound buffer (8 MiB per socket), and inbound message size (1 MiB, text and binary, per frame and per fragmented total).

The inbound cap is enforced at frame-read time by fastwebsockets, via a workspace patch to robtfm/fastwebsockets `0_8_hotfix` (0.8.1 + upstream denoland/fastwebsockets@bc39c1f, the unreleased fragmented-total cap, + a 1 MiB default). The pinned deno_websocket exposes no `set_max_message_size` hook, so the fork default is the cap; an oversized message errors and closes the socket without the payload being buffered, surfacing to the scene as ordinary error + close events. deno itself is unchanged.

Verified: `cargo clippy --all -- -D warnings` and `cargo fmt --all --check` clean. No tests here; the fragmented-cap behaviour is covered by the tests shipped with the upstream fastwebsockets commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4578,8 +4578,7 @@ dependencies = [
 [[package]]
 name = "fastwebsockets"
 version = "0.8.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9dac026e15fb7e44d768880b868a0fd5bd30ffdee272e88b3060f657a5a72947"
+source = "git+https://github.com/robtfm/fastwebsockets?branch=0_8_hotfix#1db2dbc2d1db519e6ba76741fd76f13519ab76f7"
 dependencies = [
  "base64 0.21.7",
  "bytes",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -420,6 +420,10 @@ deno_url = { git = "https://github.com/robtfm/deno", branch = "1_46_hotfix" }
 deno_webidl = { git = "https://github.com/robtfm/deno", branch = "1_46_hotfix" }
 deno_web = { git = "https://github.com/robtfm/deno", branch = "1_46_hotfix" }
 deno_websocket = { git = "https://github.com/robtfm/deno", branch = "1_46_hotfix" }
+# 0.8.1 + upstream bc39c1f (fragmented-message size cap, unreleased) + 1 MiB default cap.
+# deno_websocket's pinned API exposes no set_max_message_size hook, so the fork default is
+# how scene websocket inbound message size is bounded.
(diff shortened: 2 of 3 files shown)


robtfm:
Updated this branch (with maintainer edits) — summary of what changed and why:

**Rebased onto `main` and retargeted** — the headless-storage prerequisites this was stacked on have landed, so the PR is now the single cap commit plus a rework commit.

**Slot tracking restructured.** The drop-guard resources parked in the deno resource table didn't buy anything: every release path was already explicit, and on isolate teardown the counter died together with the guards, so the Drop was never load-bearing. Open sockets are now a plain rid `HashSet` plus an in-flight `connecting` counter in the per-scene OpState. The reserve-before-handshake property is kept (a synchronous `new WebSocket()` burst can't race past the cap), and the slot is freed on `op_ws_close`, on the glue's error path (kind 3 — the one teardown path that never calls `op_ws_close`; close codes always do), or with the isolate.

**Inbound cap moved out of the op layer.** The post-hoc check in `op_ws_get_buffer_as_string` had three problems: it only rejected a text message *after* the full payload had been buffered (so the memory spike still happened), binary was uncapped entirely, and the `take_any` force-close removed the ws resource while the JS glue still reported OPEN — a subsequent `send()` would hit the raw fns' unwrapped resource lookups and panic the shared sidecar. It's replaced by enforcement inside fastwebsockets at frame-read time, via a workspace `[patch.crates-io]` to `robtfm/fastwebsockets` `0_8_hotfix`: the 0.8.1 release plus upstream denoland/fastwebsockets@bc39c1f (caps fragmented totals — upstream's `max_message_size` was per-frame only and unreleased anywhere, so fragmentation bypassed it) plus a 1 MiB default. The pinned deno_websocket has no `set_max_message_size` hook, so the fork default is the cap; deno itself is untouched. Oversized messages (text or binary, single-frame or fragmented) now error before the payload is buffered and surface to the scene as normal error + close events.

Net: same four caps as before (32 sockets / 15s handshake / 8 MiB outbound / 1 MiB inbound), with the binary gap and the disclosed residual closed, and ~50 fewer lines in the override layer.

PR #1033, opened 2026-07-30
scene sandbox: bound comms rate, size and bus queue

eordano:
Per-scene comms has no inbound rate or size cap, no outbound amplification bound, and an unbounded message-bus queue. One scene can flood the bus and the cost lands on the shared process.

Adds an inbound rate and size cap, an outbound amplification bound, and a bounded bus queue.

Note on the size cap: LiveKit's own negotiated data-message maximum is 64KiB, stricter than the 128KiB here, so in practice the rate cap is the effective inbound protection. The size mechanism was proven by lowering it to 32KiB rather than by hitting 128KiB.

Verified: `cargo check --workspace` clean on this base.

### DIFF
--- a/crates/common/src/rpc/stream_sender.rs
+++ b/crates/common/src/rpc/stream_sender.rs
@@ -1,8 +1,20 @@
-use std::sync::{Arc, Mutex};
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc, Mutex,
+};
 
 use crate::rpc::*;
 use serde::{de::DeserializeOwned, Deserialize, Deserializer, Serialize};
 use tokio_util::sync::CancellationToken;
+
+// In-flight counter for a bounded stream: the sender drops when full, the receiver decrements as it drains.
+#[derive(Debug)]
+struct QueueGaugeInner {
+    len: AtomicUsize,
+    cap: usize,
+}
+
+type QueueGauge = Arc<QueueGaugeInner>;
 
 #[derive(Clone)]
 pub enum LocalChannel<T> {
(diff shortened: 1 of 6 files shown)


robtfm:
Rebased onto main (the change is client-generic, nothing here depends on the headless-storage branch) and pushed four follow-up commits on top:

- **Bounded channels instead of the QueueGauge side-counter** — tokio's bounded mpsc with `try_send` gives drop-when-full natively, so the manual in-flight accounting (and its check-then-act race) is gone. All `RpcStreamSender` channels are now bounded: 4096 default for engine-fed streams, 1024 for the peer-fed bus channels. Drops should never happen in normal operation, so a drop logs a warning with a preview of the lost message.
- **Outbound budget enforced in the ops, advisory JS caps dropped** — scene code can reach `Deno.core.ops` directly, so the caps in the CommunicationsController wrapper were bypassable. Replaced with a 512 messages-per-tick budget in the send ops themselves (reset in the shared per-tick flush); the wrapper reverts to a plain pass-through. The per-call peer cap is subsumed, since 512 sends/tick bounds distinct recipients regardless.
- **Inbound size cap removed** — LiveKit's negotiated 64KiB data-message maximum is stricter than the 128KiB checked here and `participant_payload` only sees LiveKit traffic, so the check could never fire (as the PR description anticipated).
- **Rate-limiter eviction tied to participant lifecycle** — an `OnRemove` observer on `LivekitParticipant` evicts entries on both explicit disconnect and relationship-cascade despawn on room teardown (which never fires `ParticipantDisconnected`). That bounds the map by connected-participant count, so the `MAX_RATE_ENTRIES` backstop and its inline clearance are removed.

Each commit compiles standalone; the tip passes fmt, clippy `-D warnings`, and the wasm build.

PR #1038, opened 2026-07-31
fix/ui/09: a teleport to a place carries its realm (Places/map jump-in from a World, /goto main)

pabloes:
Stacked on #1023 (`fix/ui/08-minimap`).

## What

- **Places/map jump-in from a World now carries realm + parcel together** (`teleportToPlace`), instead of the two independent `changeRealm`/`teleport` actions racing each other.
- **`/goto main` / `/goto genesis [x,y]`** now name Genesis Plaza (defaulting to parcel `-3,-2`) instead of only changing realm. `/goto <world> [x,y]` gained the same optional trailing-parcel override, for testing.
- **Jump-in icon fixed**: `JumpInGlyph()` drew a pencil, not an arrow — a copy/paste-wrong-glyph mistake duplicated identically in `PlaceCard.tsx` and `FeaturedCard.tsx`. Replaced with an arrow-into-a-doorframe glyph in both.
- **Places toolbar** (Explore all / Favourites / My places + search + category + sort) now renders above the Live Now / Featured sections instead of below, so it's the first thing on the page.

## Why

- **The jump-in bug** (reported by Gon): from inside a World, jumping into a Genesis City place from **Places** teleports but never leaves the World, so the destination coordinates address that World's own parcel grid instead of Genesis's. `/goto main` does change realm, hence "it works with the command but not from the page." This has been broken since the react-web port (`6abf0ba7`) — not a recent regression. The old system-scene HUD (`community-place-popup.tsx`) and the native explorer (`system_ui/src/discover.rs`) both fire a realm change + teleport together; react-web never did. The map's JUMP IN had the same hole.
- **The jump-in icon**: visible on first impression, so worth fixing on its own regardless of the realm work above.

## How

A place is modelled as what it is, **realm + parcel**, and travels as one `teleportToPlace` request. Plain `teleport` keeps its old meaning: a parcel of the realm you're already in (chat location links, a photo's capture spot).

Resolution lives in the **bridge scene**, not React, because that's where the current realm is known first-hand — React only sees the 2s `realmInfo` poll and would read a stale realm right after a change. The scene skips the switch when the player is already in the target realm, since a `changeRealm` to the realm you're in is a full scene purge and reconnect, not a no-op.

The parcel can't ride on `changeRealm`'s promise: that promise resolves as soon as the engine *accepts* the change (`restricted_actions/src/lib.rs` queues a `ChangeRealmEvent` and answers `Ok` immediately) — the about-fetch, scene purge, and pointer re-resolve all happen over the following frames. Applying the parcel then lands on the realm being *left*, which showed up as two loading screens (one out-of-world round trip in the old world, then another once the new realm lands). So the parcel is held until `getRealm` reports the target realm, with the realm poll tightened to ~10×/s while one is pending and a 60s drop-dead.

`/goto genesis` defaults to parcel `-3,-2` (the plaza's base parcel), **deliberately not `0,0`**: a bare realm change from a World lands you on whatever Genesis parcel matches the world coordinates you happened to be at (`RealmInitialLocation::Base`), and `0,0` also happens to be where a World's own scene usually sits. Because the engine only purges scenes for realms with `scenes_urn`, that World scene survives the realm change and is still "ready" — so a teleport to `0,0` finds it and drops the player in before Genesis Plaza has even spawned (no loading screen, empty land). `-3,-2` sidesteps that; the landing spot is identical either way since the engine spawns at the scene's own spawn points. `/goto genesis x,y` overrides the default for testing other parcels.

**Related, not blocking:** #1042 fixes the actual root cause engine-side (`handle_out_of_world` was accepting a `ScenePointers` entry from the wrong realm). Its own description calls this out directly: *"Once this ships, react-web can drop a workaround (`/goto main` currently avoids parcel `0,0`...)"*. This PR doesn't depend on #1042 to merge — the `-3,-2` workaround is correct standalone — but once #1042 ships, that special-case can simplify back to `0,0`.

## Deliberately unchanged

Chat location links, the gallery's Jump In, and `/goto x,y`: those coordinates are relative to the realm you're already in, and the old HUD didn't change realm for them either.

## Testing

`tsc --noEmit` clean in both `react-web` and `bridge-scene`. Full unit suite green (49 files, 334 tests), including coverage for the message contract (`world.test.tsx`, `chat.test.tsx`) and the map's JUMP IN. Runtime confirmation of the realm-arrival wait still wants a pass with the engine plus a rebuilt bridge scene, entering from a World.

## Anything Else

Backlog item 44, engine-side: the loading screen closes before the destination scene has actually loaded (empty parcel for a few seconds, pending-asset counter never appears). The asset gate is tick-bounded rather than completion-bounded — `gltf_container.rs` only blocks while `tick_number <= 5`, then forces `GltfLoadingCount` to 0. Needs a WASM rebuild, so it's not part of this stack; one to discuss with @robtfm.

### DIFF
--- a/react-web/bridge-scene/src/domains/world.ts
+++ b/react-web/bridge-scene/src/domains/world.ts
@@ -64,29 +64,20 @@ export function registerWorld(ctx: Ctx): void {
     })
   }
 
-  // A place teleport held until its realm change lands (see the teleportToPlace handler).
-  // `waited` is seconds accumulated by the realm poll, so a realm that never arrives drops it
-  // instead of teleporting into whatever realm the player ends up in.
-  let pendingPlace: { realm: string; x: number; y: number; waited: number } | null = null
-  const PLACE_ARRIVAL_TIMEOUT = 60
+  // A teleport held until its realm change lands (see the teleport handler). `waited` is seconds
+  // accumulated by the realm poll, so a realm that never arrives drops it instead of teleporting
+  // into whatever realm the player ends up in.
+  let pendingTeleport: { realm: string; x: number; y: number; waited: number } | null = null
+  const REALM_ARRIVAL_TIMEOUT = 60
 
-  ctx.on('teleport', (msg) => {
-    teleportToParcel(msg.x, msg.y)
-  })
-
-  // Travel to a world/realm (e.g. boedo.dcl.eth). The engine auto-grants ChangeRealm for our
-  // super-user scene, so the React HUD owns the confirmation prompt.
-  ctx.on('changeRealm', (msg) => {
-    changeRealm({ realm: msg.realm }).catch((e: unknown) => {
-      console.error('[world] changeRealm failed', e)
-    })
-  })
-
-  // A place is realm + parcel. Resolved here rather than in React because this is where the
-  // current realm is known first-hand: React only sees the 2s realmInfo poll, and a teleport
-  // fired right after a realm change would read a stale one. The realm switch is skipped when the
-  // player is already there — a changeRealm to the realm you are in is a full scene purge +
-  // reconnect, not a no-op.
+  // Teleport to a parcel. Without a realm the coordinates are relative to wherever the player
+  // already is (a chat location link, a photo's capture spot) and nothing changes realm.
+  //
+  // With one, the parcel belongs to that realm, and getting there is part of the trip. It is
+  // resolved here rather than in React because this is where the current realm is known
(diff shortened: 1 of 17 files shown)


pabloes:
waiting for https://github.com/decentraland/bevy-explorer/pull/1042 in order to fix all known teleport/jump issues before opening and requesting review to merge

robtfm:
closed in favour of 1227

PR #1053, opened 2026-08-04
DCL Cast and presentations

hukasu:
This broke volume control on VideoPlayers, though

## PR TL;DR
Previosly the control of streams were done in `comms` with the `Streamer`, `StreamBroadcast`, and `StreamViewer` components, this does not fit well with the correct behavior since you need to keep track of priorities between different streams, so this logic was pulled into a new crate called `livestream_manager` and the livekit video tracks now take a component based on their priority
* `TransmitterKind::Presentation` for tracks by participants with the `presentation-bot:` prefix
* `TransmitterKind::Screenshare` for any other track with `TrackSource::Screenshare`
* `TrasmitterKind::VideoCast` for tracks with `TrackSource::Camera` and participant identity starting with `stream:`
* `TransmitterKind::Stream` for tracks with `TrackSource::Camera` and participant identity ending in `-streamer`
* `ActiveVideoCast` is added to `TransmitterKind::VideoCast` that are also in the rooms `active_speaker`, and removed after 3 seconds of being removed from `active_speakers`

`av` had a huge refactor, `AudioStream` and `VideoPlayer` now rely more heavly on observers


crates/av/src/html_video_player.rs
```diff
@@ -478,254 +459,234 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
+        Some(texture) => texture.0.clone(),
+    };
+
+    debug!(
+        "Creating new html media entity for {} targeting \"{}\"",
+        entity,
+        &(**player_source)
+    );
+    match &(**player_source) {
+        "" | LIVEKIT_VIDEO_STREAM => (),
```
robtfm: **DCL Cast cannot activate on web.** `Stream` is now only inserted in the native `new_player_source` (`video_player.rs:102`); nothing on the wasm path inserts it. Without `Stream`, `stream_on_add`/`should_be_playing_on_add` in `lib.rs` never insert `ActiveReceiver`, so `ReceiverImage` never arrives and the whole `receiver_image_added`/`WaitingForStream` machinery below is dead code. The previous revision inserted `Stream` platform-independently in `av_player_on_insert` — this restructure lost that for wasm.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +459,234 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
 ) {
-    let scene_volume = audio_settings.scene();
-    for (ent, container, player, maybe_texture) in av_players.iter() {
-        let Ok(context) = scenes.get(container.root) else {
-            continue;
-        };
+    let entity = trigger.target();
+    let Ok((config, mut html_media_entity, has_should_be_playing)) = av_players.get_mut(entity)
+    else {
+        unreachable!("Infallible query");
```
robtfm, line 254: This `unreachable!` is reachable: `new_player_source` creates no `HtmlMediaEntity` for `""` or `LIVEKIT_VIDEO_STREAM` sources (and for stream sources it also doesn't exist while `WaitingForStream`), but `T::Config` is always inserted by `av_player_on_insert`. Any web scene with an empty-src or cast VideoPlayer panics here. Needs the same graceful-return treatment the native observers got in 475cdc00.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +459,234 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
-                    image_handle.clone(),
-                )
-            };
+fn player_position_added<T: AVPlayer>(
+    trigger: Trigger<OnInsert, T::Position>,
+    mut av_players: Query<(&T::Position, &mut HtmlMediaEntity<T>)>,
+) {
+    let entity = trigger.target();
+    let Ok((position, mut html_media_entity)) = av_players.get_mut(entity) else {
+        unreachable!("Infallible query");
```
robtfm, line 285: Same as `player_config_added`: reachable for `""`/stream sources — needs a graceful return instead of `unreachable!`.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +459,234 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
+    };
 
-            let video_volume = player.volume();
-            video.set_loop(player.r#loop());
-            video.set_volume(video_volume * scene_volume);
-            let video_output = VideoTextureOutput(image_handle);
+    debug!("Seeking AVPlayer to {}", **position);
+    html_media_entity.current_time = **position;
```
robtfm: This only updates the bookkeeping field — it never seeks the media element (`self.media.set_current_time(...)`), and it corrupts the offset reported in subsequent `PbVideoEvent`s.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +459,234 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
-                        | TextureUsages::TEXTURE_BINDING
-                        | TextureUsages::RENDER_ATTACHMENT;
-                    image.transfer_priority = bevy::asset::RenderAssetTransferPriority::Immediate;
-                    image.data = None;
-                    images.add(image)
-                }
-                Some(texture) => texture.0.clone(),
-            };
+    if config.playing() && has_should_be_playing {
+        html_media_entity.play();
```
robtfm, line 270: `play()` is now called once from the observers; the old per-frame path retried when the element became ready. If the browser rejects the initial `play()` (autoplay policy — common before the first user gesture), the player never starts.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +459,234 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
+    } else {
+        html_media_entity.stop();
+    }
+    html_media_entity.set_volume(config.volume());
```
robtfm: Missing the `* audio_settings.scene()` multiplier. `update_html_video_player_volumes` only re-applies it `run_if(resource_exists_and_changed::<AudioSettings>)`, so every scene config update resets the volume to unscaled until the user next touches audio settings.

crates/av/src/lib.rs
```diff
@@ -357,14 +636,20 @@ fn video_player_should_be_playing(
...
-                        };
+            |(
+                ent,
+                player_source,
+                player_config,
+                has_in_scene,
+                has_should_be_playing,
+                transform,
+            )| {
+                if player_config.playing() {
```
robtfm: **Pausing never removes `ShouldBePlaying<VideoPlayer>`.** Non-playing entities are filtered out of `sorted_players` here, so neither removal loop below ever sees them. A stream player that pauses keeps `ShouldBePlaying` → keeps `ActiveReceiver` → the track stays subscribed and cast audio keeps playing after the scene pauses its screen. `audio_stream_should_be_playing` handles exactly this case (`(_, true, false)` → remove); the video variant needs the same.

crates/av/src/lib.rs
```diff
@@ -390,132 +675,49 @@ fn video_player_should_be_playing(
...
-                context.update_crdt(
-                    SceneComponentId::VIDEO_EVENT,
-                    CrdtType::GO_ANY,
-                    container_entity.container_id,
-                    &event,
-                );
-            }
-        }
+fn video_texture_output_image_changed(video_texture_outputs: Populated<&mut VideoTextureOutput>) {
+    // TODO make it so that this workaround isn't needed
```
robtfm: Still flagging this from the first pass: force-marking every `VideoTextureOutput` changed each `FixedUpdate` tick defeats change detection for all video materials, not just streams. Worth resolving the underlying issue before merge, or at least scoping it to the shared stream image.

crates/av/src/video_player.rs
```diff
@@ -73,91 +60,203 @@ fn init_ffmpeg() {
...
+    };
+
+    debug!(
+        "Creating new sinks for {} targeting \"{}\"",
+        entity,
+        &(**source)
+    );
+    let (video_sink, audio_sink) = match &(**source) {
+        "" => noop_sinks((**source).to_owned(), create_image_handle(), 1.),
+        LIVEKIT_VIDEO_STREAM => return,
```
robtfm: Since stream players get no sinks and `play_videos` only iterates `AVSinks<VideoPlayer>`, native stream players now emit no `PbVideoEvent` at all — the old `stream_should_be_played`/`streamer_joined` paths sent `VsLoading`/`VsPlaying`/`VsPaused` and nothing replaces them. Scenes that show "stream live/offline" UI off video events lose that on native.

crates/livestream_manager/src/plugin.rs
```diff
@@ -0,0 +1,445 @@
...
+            entity_cmd.insert(AudioStreamer(*livestream_manager));
+            if *transmission_kind == TransmissionKind::Stream {
+                entity_cmd.insert(ActiveAudioTransmitter);
+            }
+        }
+    }
+}
+
+fn audio_transmitter_kind_on_replace(
+    trigger: Trigger<OnReplace, TransmitterKind>,
```
robtfm: Copy-paste: this listens on `OnReplace` of **`TransmitterKind`** (the video kind) but cleans up the audio components. Audio-kind replacement/removal cleans nothing (masked today by despawn-driven relationship cleanup), and a video-kind replace targets audio components its entity doesn't have. Should be `Trigger<OnReplace, AudioTransmitterKind>`.

crates/livestream_manager/src/plugin.rs
```diff
@@ -0,0 +1,445 @@
...
+    let Ok(transmitter_kind) = transmitter.get(entity) else {
+        unreachable!("Infallible query");
+    };
+
+    let mut entity_cmd = commands.entity(trigger.target());
+
+    match transmitter_kind {
+        AudioTransmitterKind::Cast => {
+            entity_cmd.insert(AudioCaster(*livestream_manager));
+            if *transmission_kind == TransmissionKind::Cast {
```
robtfm: This gates on `TransmissionKind`, which is `Cast` whenever any cast transmitter exists — regardless of receivers. An audio track published while nobody is watching gets `ActiveAudioTransmitter` → subscribed and audible with zero receivers, and since `OnExit(TransmissionState::Cast)` never fires (the state never entered `Cast`), nothing removes it. Should check the receiver-aware `TransmissionState` (same for the `Stream` arm below).

crates/comms/src/livekit/participant/plugin.rs
```diff
@@ -317,134 +292,6 @@ fn participant_metadata_changed(
...
-        for livekit_track in tracks.iter_many(publishing.collection()) {
-            let track = livekit_track.clone();
-            livekit_runtime.spawn(async move {
-                track.set_subscribed(false);
-            });
-        }
-    }
-}
-
 fn change_volume_of_tracks(
```
robtfm: This chain is now severed at both ends: nothing triggers `ChangeVolume` (the av players' producers were deleted) and nothing consumes `TrackVolume` (`update_track_volume`/`update_tracks_volume` were deleted). Functionally that means a scene's `VideoPlayer.volume` no longer affects stream/cast audio at all — the only remaining control is the global scene-volume setting. Either reinstate the path or remove `ChangeVolume`/`TrackVolume`/this observer as dead code.

crates/comms/src/livekit/participant/plugin.rs
```diff
@@ -575,25 +441,50 @@ fn is_no_longer_speaking(
...
+    active_speakers: Populated<(Entity, &ActiveSpeakerGracePeriod)>,
+    time: Res<Time<Real>>,
+) {
+    const GRACE_PERIOD: f64 = 3.;
+
+    let elapsed = time.elapsed_secs_wrapped_f64();
+    let wrap_period = time.wrap_period().as_secs_f64();
+    for (entity, active_speaker_grace_period) in active_speakers.into_inner() {
+        let mut active_speaker_grace_period = **active_speaker_grace_period;
+        if active_speaker_grace_period >= wrap_period - GRACE_PERIOD {
```
robtfm: Nitpick: this wrap adjustment only covers stamps within `GRACE_PERIOD` of the wrap boundary. A stamp taken mid-range just before `elapsed` wraps makes `elapsed - stamp` hugely negative, so `ActiveVideoCast` sticks until elapsed catches back up (up to a full wrap period). Probably acceptable as an hour-scale edge case, but a simple `Timer`/`Duration`-since-stamp would avoid it.

robtfm (commented):
> [!NOTE]
> This review was generated by Claude Code (posted on Rob's behalf).

Re-review of the updated branch. The native-side issues from the previous pass are genuinely addressed — the `sinks.audio` copy-paste dispose bug, the native `unreachable!`s, the receiver lifecycle (`ActiveReceiver` now correctly gated on `Stream` + `ShouldBePlaying`), the unconditional bot-audio subscription, the camera→`VideoStream` classification (and the `-streamer` convention is back), native URL-player volume, and the log spam. The active-speaker grace period and the test suites are welcome additions.

The blocker is that **`html_video_player.rs` wasn't touched**, and one restructure regressed the web path relative to the earlier revision of this PR: `Stream` is now only inserted in the native `new_player_source`, so on wasm the feature can't activate at all — and the guaranteed `unreachable!` panics for empty-src/stream players are still present. Details inline.

Also still open from the first pass: native stream players emit no `PbVideoEvent` at all (`VsLoading`/`VsPlaying`/`VsPaused` senders were deleted and `play_videos` only iterates ffmpeg sinks), and on wasm `update_av_players` is filtered `With<ShouldBePlaying>` so paused players stop reporting events. Scenes that drive UI off video events for streams will break.

crates/av/src/html_video_player.rs
hukasu: isn't this only an issue when opening a new tab that already has a `<video>` on it? with the usual flow of the app, the `<video>` will only ever appear long after input is received

crates/av/src/lib.rs
```diff
@@ -271,25 +437,139 @@ impl Plugin for AVPlayerPlugin {
...
+
+#[expect(clippy::type_complexity)]
+fn stream_on_add<T: AVPlayer>(
+    trigger: Trigger<OnAdd, Stream>,
+    mut commands: Commands,
+    av_players: Query<(Has<ShouldBePlaying<T>>, Option<&T::Config>), (With<T>, With<Stream>)>,
+) {
+    let entity = trigger.target();
+    let Ok((has_should_be_playing, maybe_config)) = av_players.get(entity) else {
+        unreachable!("Infallible query");
```
robtfm, line 549: **Scene-triggerable panic.** `new_player_source` is generic over the player type and inserts `Stream` for any `AVPlayer` whose source is the cast URL — including `AudioStream`. `Trigger<OnAdd, Stream>` isn't typed by `T`, so this observer also fires for that `AudioStream` entity, the `With<VideoPlayer>` query misses, and this `unreachable!` panics — all builds, native and wasm. Repro: spawn `AudioStream { url: LIVEKIT_VIDEO_STREAM }` with a `ContainerEntity` — panics right here (on main the same component just failed gracefully through ffmpeg). Needs a graceful return (same for `should_be_playing_on_add` below), or only inserting `Stream` on `VideoPlayer` entities.

crates/av/src/lib.rs
```diff
@@ -271,25 +437,139 @@ impl Plugin for AVPlayerPlugin {
...
+    };
+
+    if has_should_be_playing {
+        commands.entity(entity).insert((
+            ActiveReceiver,
+            ReceiverVolume(maybe_config.map(|config| config.volume()).unwrap_or(0.)),
```
robtfm, line 556: `unwrap_or(0.)` starts the receiver muted whenever the config isn't present yet, and disagrees with `ReceiverVolume::default()` (1.0). Presumably should be `1.` — same in `should_be_playing_on_add` below.

crates/av/src/lib.rs
```diff
@@ -341,181 +618,411 @@ fn audio_stream_should_be_playing(
...
+}
+
+impl PartialOrd for ShouldBePlayingCandidate {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for ShouldBePlayingCandidate {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
```
robtfm, line 678: **Equidistant players get dropped from scheduling.** This `Ord` compares only `(in_scene, distance)` while the derived `Eq` compares all fields, and `BTreeSet` goes purely by `Ord` — so a second candidate at the same distance is silently not inserted into the set: it never gains `ShouldBePlaying`, and never hits either removal loop. Exactly-equal float distances are real — symmetric layouts (screens at (5,0,0)/(−5,0,0) with the player between them). Verified: two such screens with `max_videos: 4` and the second never plays. Adding `entity` as a final tie-break fixes it.

Side note: the tests in this module never propagate transforms, so every `GlobalTransform` is identity and all distances compare equal — `none_initialy_with_should_be_playing` currently passes because of this dedup, not because of distance ordering.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +487,311 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
+        has_stream,
+        maybe_receiver_volume,
+    )) = av_players.get_mut(entity)
+    else {
+        unreachable!("Infallible query");
+    };
+    if has_stream {
+        if let Some(mut receiver_volume) = maybe_receiver_volume {
+            debug!("Updated volume of stream.");
+            **receiver_volume = config.volume() * audio_settings.scene();
```
robtfm: **Scene volume is applied twice on wasm.** This writes `volume * scene()` into `ReceiverVolume`, and the wasm `update_track_volume` multiplies by `audio_settings.scene()` again — so cast audio is `v·s²` after any config update, and `v·s` right after stream start (the initial `ReceiverVolume` insert in `av/lib.rs` writes the plain volume, so the level also jumps the first time the scene touches the config). Native is consistent: plain per-track volume × `scene()` on the `LivekitAudioManager` main track. Write plain `config.volume()` here and let `update_track_volume` own the settings multiplier.

crates/av/src/html_video_player.rs
```diff
@@ -478,254 +487,311 @@ impl<T: AVPlayer> Drop for HtmlMediaEntity<T> {
...
     mut av_players: Query<
-        (
-            Entity,
-            &ContainerEntity,
-            Option<&mut HtmlMediaEntity<T>>,
-            Has<ShouldBePlaying<T>>,
-        ),
-        With<T>,
+        (Entity, &ContainerEntity, Option<&mut HtmlMediaEntity<T>>),
+        (With<T>, With<ShouldBePlaying<T>>),
```
robtfm: main's version of this system iterated all players with `Has<ShouldBePlaying<T>>`; filtering on `With<ShouldBePlaying<T>>` means a paused player is never visited again — the element goes `VsPaused` after `stop()`, but the state change is never reported, so the scene never receives a `VsPaused` video event (and a pause→resume cycle emits no state events at all, since `last_state` is still `VsPlaying` on resume). Native `play_videos` still reports these. Please restore the old shape — iterate all players and keep `ShouldBePlaying` as data.

crates/av/src/html_video_player.rs
```diff
@@ -839,11 +905,59 @@ fn perform_video_copies(
...
+struct WaitingForStream;
+
+fn receiver_image_added(
+    trigger: Trigger<OnAdd, ReceiverImage>,
+    mut commands: Commands,
+    video_players: Query<&ReceiverImage, With<VideoPlayer>>,
+) {
+    let entity = trigger.target();
+    let Ok(receiver_image) = video_players.get(entity) else {
+        unreachable!("Infallible query");
```
robtfm: The `av/lib.rs` version of this observer guards with `if let Ok`; this one panics if `ReceiverImage` ever lands on a non-`VideoPlayer` entity (reachable today via the `AudioStream` + cast-URL path above). Make it a graceful return too.

crates/av/src/html_video_player.rs
```diff
@@ -839,11 +905,59 @@ fn perform_video_copies(
...
+    };
+
+    if let Some(html_media_entity) = HtmlMediaEntity::<VideoPlayer>::new_stream(
```
robtfm: **Transmitter switches freeze the stream on web.** `new_stream` wraps whatever `<video>` is at index 0 of the container at acquisition time, and nothing ever re-acquires: `livekit_web_bindings.js` never detaches a *video* track's element on `TrackUnsubscribed` (only audio), and each newly-subscribed video track appends a new element — so after the first `ActiveTransmitter` switch (which is every active-speaker change during a cast), the wrapped element is dead and index 0 stays pointing at it. Unsubscribing fires no error event, so the `VsError` retry in `update_av_players` never triggers; the viewer just freezes on the last frame. Needs the JS to detach/remove video elements on unsubscribe, plus re-acquiring the element when the subscribed track changes (or looking it up by track sid instead of index 0). Native is fine — `copy_frame` pipes whichever transmitter is active into the shared image.

crates/av/src/html_video_player.rs
```diff
@@ -839,11 +905,59 @@ fn perform_video_copies(
...
+        LIVEKIT_VIDEO_STREAM.to_owned(),
+        (*receiver_image).clone(),
+    ) {
+        commands.entity(entity).insert((
```
robtfm: **Multiple stream receivers fight over the one shared element.** Every stream `VideoPlayer` gets its own `HtmlMediaEntity` wrapping the same `<video>`: one player pausing (scene config, or `max_videos` eviction) calls `stop()` on the element and freezes the others; `new_stream`'s `stop()`/`play()` hack un-pauses it regardless of whether anyone should be playing; and `Drop` calls `media.remove()`, deleting the JS-owned element from the DOM when any single stream player despawns or changes source — the surviving receivers freeze with no recovery until receiver count hits zero and a resubscribe re-attaches. The element wants a single owner — e.g. one manager-owned wrapper keyed by track, with per-player entities only holding the image — rather than per-receiver ownership of the shared element.

robtfm (commented):
Re-review against the last round. Confirmed fixed: wasm `Stream` insertion, the reachable unreachables on the html observers, seek, the scene-volume multiplier on config change, pause removing `ShouldBePlaying`, the `set_changed` scoping, the audio-kind `OnReplace` cleanup, the receiver-aware audio gating (nice test coverage on that), and the grace-period timer. The state machine and its tests are looking solid.

Still open from last round: the autoplay retry and native stream video events — replied in those threads.

The comments below are issues in the code added since then (or surfaced by the new wasm flow). The two `av/lib.rs` ones (cast-URL `AudioStream` panic, equidistant-player dropout) are confirmed by running minimal repros against this branch. The two stream-element lifecycle ones lean on `livekit_web_bindings.js` behaviour that predates this PR, but the multi-transmitter model here is what makes it bite.

crates/av/src/html_video_player.rs
robtfm: That covers the usual interactive flow, but a reload with a persisted login (or a deep link) boots the engine with zero gestures, and a rejected play() currently has no retry at all — recovery only happens if the scene happens to re-send the config. Since update_av_players already visits every ShouldBePlaying player each frame, retrying play() there when config.playing() but the state is stuck at VsReady/VsPaused should close this cheaply — please add that if it's straightforward.

crates/av/src/video_player.rs
robtfm: Still open in the current revision — play_videos only iterates AVSinks and stream players still return early with no sinks, so native cast/stream players emit no video events at all (wasm reports via the HtmlMediaEntity path). Fine to defer to a follow-up if you'd rather land this first — but please open an issue if so, flagging in case it just got missed.

crates/av/src/lib.rs
robtfm: Following up after tracing what this heartbeat actually costs: it's not just change-detection noise. Each `set_changed` makes the `update_materials` tail loop stamp every mesh sourcing the stream texture with `RetryMaterial`, and video-textured materials are deliberately uncacheable (`can_cache = false` — they carry a `source_entity`), so each retry goes through `materials.add` and allocates a brand-new `SceneMaterial` asset. Net effect while a stream is on screen: a new material asset, a new bind group, and a pipeline respecialization/re-queue for every stream-screen mesh, every FixedUpdate tick (64Hz), plus the corresponding asset drop — bypassing the material caching that path was specifically optimised around.

The legitimate invalidation points are discrete: the placeholder→sized image swap, stream acquisition, and transmitter/track changes — and those all go through `VideoTextureOutput` inserts, which already mark it changed naturally. Once the image is sized, the frame copies mutate the texture in place and no material rebuild should be needed at all. Please replace this heartbeat with those explicit invalidations and delete the FixedUpdate system — if there's a residual case where the screen goes stale without it (e.g. a resize recreating the GpuImage under a live bind group), let's pin that case down and invalidate on that event specifically rather than every tick.

crates/av/src/lib.rs
hukasu: This is because i rather have the audio muted for a frame then playing when config becomes available and not muted, than have it briefly play then mute when config becomes available and muted

robtfm (commented):
Fourth pass, against `f6d72a0b`. The big items from last round are in good shape: the stream element lifecycle on web is properly solved by wrapping the track's own `videoElement` on the track entity and copying into the manager image, the 64Hz `set_changed` heartbeat is gone in favour of discrete `TransmissionUpdated` invalidations, the Ord tie-break / AudioStream panic / wasm volume / `Has<ShouldBePlaying>` fixes all check out, and the tests look good (av 17/17, livestream_manager 24/24, media 1/1 locally with `--features ffmpeg`; fmt and clippy clean on those crates). The `unwrap_or(0.)` reasoning is fine by me.

The `SetState` commit (`6426bf75`) needs another look though — it introduces a scene-unload panic and two event-correctness problems, detailed inline. Summary:

1. **Blocker:** the deferred `trigger(SetState { VsNone })` in `av_player_on_remove` panics when the entity is being despawned (scene unload), and never actually delivers `VsNone` on plain component removal.
2. `receiver_image_updated` touches every `VideoTextureOutput`, so every regular video player gets a spurious `VsPlaying` (and a material realloc) on each transmitter change.
3. Stream state events are still incomplete: receivers that join a running cast never get `VsPlaying`, and nothing is sent when the transmission ends.
4. The autoplay retry doesn't cover `VsReady`, which is where a rejected `play()` leaves the element.

Plus a few small ones. Once 1–3 are sorted I think this is close.

crates/av/src/lib.rs
```diff
@@ -271,25 +421,183 @@ impl Plugin for AVPlayerPlugin {
...
+        T::Config,
+        T::Position,
+        InScene,
+        ShouldBePlaying<T>,
+        Stream,
+    )>();
+    #[cfg(feature = "ffmpeg")]
+    commands.entity(entity).try_remove::<AVSinks<T>>();
+
+    commands.entity(entity).trigger(SetState::<T> {
```
robtfm: **Panics on scene unload.** This `trigger` is a deferred entity command. When the `OnRemove` fires because the entity is being *despawned* (every scene teardown with a VideoPlayer/AudioStream), the entity no longer exists by the time the command applies, and the default bevy error handler panics (the client doesn't enable `configurable_error_handler`; only headless does). Verified with a scratch test — spawn VideoPlayer + ContainerEntity, flush, `world.despawn()`, flush:

```
Encountered an error in command entity_command::trigger<av::SetState<av::VideoPlayer>>...: The entity with ID 73v1 does not exist
```

On plain component removal (the path the existing tests cover) it doesn't panic, but `set_state`'s `With<T>` query fails because the component is already gone, so it logs `Trying to set state for an entity that is not part of a scene.` and `VsNone` is never sent either — you can see two of those errors in the test log from `test_removing_*`.

`ContainerEntity` is still present inside the `OnRemove` observer, so the simplest fix is to write the `VsNone` event directly here (query `&ContainerEntity` + `&mut RendererSceneContext`) instead of deferring through `SetState`.

crates/av/src/lib.rs
```diff
@@ -341,181 +646,562 @@ fn audio_stream_should_be_playing(
...
+    debug!("ReceiverImage removed from {}", entity);
+    commands.entity(entity).try_remove::<VideoTextureOutput>();
+}
+
+fn receiver_image_updated(
     mut commands: Commands,
-    av_players: Query<(&T, &ContainerEntity, Has<StreamViewer>)>,
-    mut removed_av_players: RemovedComponents<T>,
-    mut scenes: Query<&mut RendererSceneContext>,
+    av_players: Query<(Entity, &mut VideoTextureOutput)>,
```
robtfm: This iterates every `VideoTextureOutput` in the world, not just stream receivers. So on each cast transmitter change (and each resize) every regular VideoPlayer gets `set_changed()` — which sends its material through the uncacheable-material realloc path — and, worse, a `VsPlaying` event: a scene-paused video player gets told it's playing whenever someone starts a cast. On native, AudioStreams also carry a `VideoTextureOutput` and end up in the `set_state` error log each time. Please filter `With<Stream>`.

crates/av/src/lib.rs
```diff
@@ -341,181 +646,562 @@ fn audio_stream_should_be_playing(
...
-    let Ok((av_player, container_entity)) = av_players.get(entity) else {
-        error!("ShouldBePlaying must only be added to AVPlayers.");
-        return;
-    };
 
-    if av_player.source().starts_with("livekit-video://") {
-        debug!("AVPlayer {entity} should be playing. Linking to the stream.");
+    if let Ok(receiver_image) = video_players.get(entity) {
+        debug!("ReceiverImage added to {}", entity);
         commands
```
robtfm, line 775: Nothing is sent here, and `TransmissionUpdated` only fires on transmitter change/resize — so a receiver that becomes active *after* the transmitter was chosen (i.e. walking into a scene where a cast is already running, which is the common case) only ever sees the `VsLoading` from the source insert. Nothing is sent when the transmission ends either (`ActiveTransmitter` removed via `drop_transmissions` / state exit). Scenes with live/offline UI end up: loading → maybe playing → stuck.

Suggest emitting a state here when an `ActiveTransmitter` exists, and a non-playing state on `ActiveTransmitter` removal — or, as offered last round, defer the stream-events story to an issue and keep just the parts that are correct now.

crates/av/src/html_video_player.rs
```diff
@@ -641,105 +360,102 @@ fn update_av_players<T: AVPlayer>(
...
-                                video_frame: WgpuWrapper::new(frame),
-                                target: image_id,
-                            })
-                            .report();
-
-                        av.current_time = new_time;
-                    } else {
-                        trace!("no frame (new_time == 0)");
+        if has_should_be_playing
+            && (state == VideoState::VsLoading || state == VideoState::VsBuffering)
```
robtfm: This doesn't cover the case the retry was for: a `play()` rejected by the autoplay policy leaves the element in `VsReady` (after `canplay`), which isn't in this list, so it still never starts. Please add `VsReady`. Also this currently calls `play()` every frame while loading/buffering — a cheap throttle (every N frames) would be good.

crates/av/src/html_video_player.rs
```diff
@@ -641,105 +360,102 @@ fn update_av_players<T: AVPlayer>(
...
     frame: Res<FrameCount>,
 ) {
-    for (ent, container, maybe_av, should_be_playing) in av_players.iter_mut() {
-        let Some(mut av) = maybe_av else { continue };
-
+    for (ent, container, mut av, has_should_be_playing) in av_players.iter_mut() {
         let state = av.state();
 
-        if av.source.starts_with("livekit-video://") && state == VideoState::VsError {
+        if av.source() == LIVEKIT_VIDEO_STREAM && state == VideoState::VsError {
```
robtfm: Dead now — stream players no longer own an `HtmlMediaEntity` on web (the track entity does), so `source()` is never the cast URL here.

crates/comms/src/livekit/track/plugin.rs
```diff
@@ -560,199 +566,220 @@ fn video_track_is_now_subscribed(
...
-                let Some(image) = images.get_mut(stream_image.id()) else {
-                    debug_panic!("StreamImage holds an invalid handle.");
-                };
+    let Some(video_element) = video.html_video_element() else {
+        debug!("Could not build HtmlMedia from livekit track.");
+        return;
+    };
+    let html_media =
+        HtmlMedia::video_from_element(video_element, String::new(), (*active_transmitter).clone());
+    commands.entity(entity).try_insert(HtmlMediaEntity {
```
robtfm, line 601: Small leak: this is never removed on unsubscribe (only replaced on the next `Subscribed`), so an unsubscribed track keeps its element + `requestVideoFrameCallback` loop alive until the track entity despawns. Mirror of the JS side, which still only detaches audio elements on `TrackUnsubscribed`. Removing the `HtmlMediaEntity` on `OnAdd<Unsubscribed>` (and detaching video elements in the JS handler) would keep both symmetric.

crates/av/src/lib.rs
```diff
@@ -271,25 +421,183 @@ impl Plugin for AVPlayerPlugin {
...
+    av_players: Query<(Has<ShouldBePlaying<T>>, Option<&T::Config>), (With<T>, With<Stream>)>,
+) {
+    let entity = trigger.target();
+    let Ok((has_should_be_playing, maybe_config)) = av_players.get(entity) else {
+        unreachable!("Infallible query");
+    };
+
+    if has_should_be_playing {
+        debug!("New stream {} should be playing.", entity);
+        commands.entity(entity).insert((
```
robtfm: General note for the new observers: the codebase uses `try_insert`/`try_remove` on entity commands deliberately, since a deferred command can outlive its entity — the panic above is exactly that. Worth switching the non-`try_` `insert`/`trigger` calls in these observers over.

crates/av/src/lib.rs
```diff
@@ -341,181 +659,578 @@ fn audio_stream_should_be_playing(
...
-                video_length: 0.,
-                state: VideoState::VsPlaying.into(),
-            };
-            context.update_crdt(
-                SceneComponentId::VIDEO_EVENT,
-                CrdtType::GO_ANY,
-                container_entity.container_id,
-                &event,
-            );
+            .try_insert(VideoTextureOutput((*receiver_image).clone()));
```
robtfm, line 777: `OnEnter(Transmitter::On)` covers the *first* receiver, but nothing is sent here — so a receiver that becomes active while a transmitter is already `On` only ever sees `VsLoading`. That happens whenever another receiver kept the cast alive: a portable-experience screen while you walk back into a scene whose screen re-gains `ShouldBePlaying`, or a second screen in the same scene coming into `max_videos` range. Suggest sending `VsPlaying` from here when `Res<State<Transmitter>>` is `On`.

crates/av/src/lib.rs
```diff
@@ -341,181 +659,578 @@ fn audio_stream_should_be_playing(
...
+    mut transmission_stopped: EventReader<TransmissionStopped>,
 ) {
-    let entity = trigger.target();
-    if removed_av_players.read().any(|removed| removed == entity) {
-        return;
+    if transmission_stopped.read().count() > 0 {
+        for container_entity in av_players {
+            commands.trigger(SetState::<VideoPlayer> {
+                entity: *container_entity,
+                state: VideoState::VsNone,
```
robtfm: Semantics question, not a bug: `VsNone` reads to a scene as "no video source at all", whereas the source is still set and the stream is just offline — main's old paths used `VsPaused`/`VsLoading` for that. Scenes that distinguish "offline" from "unconfigured" might care. Fine to leave if you prefer `VsNone`.

crates/comms/src/livekit/track/plugin.rs
```diff
@@ -560,199 +568,231 @@ fn video_track_is_now_subscribed(
...
-                let target_extent = Extent3d {
-                    width: frame.width().max(16),
-                    height: frame.height().max(16),
-                    depth_or_array_layers: 1,
+#[cfg(target_arch = "wasm32")]
+#[expect(clippy::type_complexity)]
+fn video_track_is_now_unsubscribed(
+    trigger: Trigger<OnReplace, Subscribed>,
+    mut commands: Commands,
+    tracks: Query<(&LivekitTrack, Has<Video>, Option<&ActiveTransmitter>), With<Subscribed>>,
```
robtfm: Nit: `tracks` is unused here (wasm `unused_variables` warning).

robtfm (approved):
Thanks — everything from the last round is fixed and verified: `SetState` no longer depends on the entity (despawn repro passes, including despawning the scene root first), `receiver_image_updated` is scoped to receivers, `VsReady` is retried, the wasm `HtmlMediaEntity` is released on unsubscribe, and the `try_*` sweep is done. Locally on `2acebfba`: av 19/19 (with the new `test_despawn_av_player`), livestream_manager 24/24, media 1/1, no more `set_state` error logs, fmt/clippy clean; CI is green including the wasm build.

Approving. Three small things inline — the first (late-joining receivers never get `VsPlaying`) is worth a one-liner before or right after merge, the other two are take-or-leave.

PR #1058, opened 2026-08-07
fix/ui/10-passport-wearables: equipped wearables + emotes in the passport

pabloes:
Adds **Equipped Wearables / Equipped Emotes** to the passport. Overview order is now About Me → Equipped → Badges.

Rebased onto `main` (previously stacked on #1038), no longer depends on it.

## What

| Surface | Change |
|---|---|
| Passport | Two new sections, 6 column grid, badge tooltips (name + tier) |
| `EquippedItemCard` | New design-system primitive (exported, in `?showcase=1`) |
| bridge-scene | `getUserProfile` resolves the equipped wearables + emotes |
| Shop | Hover reveals a SHOP deep link on collectibles |

## Why

Parity with the official passport: unity-explorer's `EquippedItems_PassportModuleController` and bevy-ui-scene's `passport-equipped-item` both render this grid, we rendered nothing. The SHOP link is the only path from someone else's outfit to buying the item.

## How

**Data.** `getUserProfile` resolves the deployed `avatar.wearables` / `avatar.emotes` through `resolveEquippedSet` plus a new `resolveEquippedEmotes` (same caching). Your **own** passport reads the live `getPlayer()` avatar instead, so a just-equipped item matches the Backpack without waiting for a redeploy and catalyst reindex. Other users have no live source, so they stay catalyst-only.

**Shop links.** `/shop/item/<contract>/<itemId>`.

- collections-v2 urns already carry the item id: built locally, no request.
- collections-v1 are slug-identified (`…:mf_sammichgamer:mf_animehair`) and the catalyst has no numeric id, so they resolve through the marketplace items API (the same one bevy-ui-scene uses), batched into one request, cached for the session.
- No listing (base wearables and emotes) means no button, not a dead link.
- Resolution is opt-in: only the passport pays for it.

**Tile.** Square rarity-gradient thumbnail, rarity-colored corner flap with the category glyph, name + rarity tag. On hover the card's `::before` extends 42px downward (the layout box never changes, so the grid does not reflow), an animated gradient ring wraps it, and the SHOP button fades in once that extension finishes.

## Verification

- `typecheck` clean (react-web + bridge-scene), `npm test` 334/334, including 8 new tier-1 tests pinning the sections, the shop-link rule, the body-shape skip and the Overview order.
- `npm run test:visual` 26/26 after regenerating the two stale baselines. `showcase.png` is unaffected: that screenshot is viewport-sized (the page scrolls in its own container), so the new showcase section falls outside it.
- Shop routes opened live: collections-v1 `0xecf0…b1/4` and collections-v2 `0xe860…89/0` both resolve to the right item page.
- Bridge-domain repro suite (P0 in both directions, self vs other emotes, shop skip/run) 6/6. Not committed: it needs vitest aliases for `@dcl/sdk/players` and `~system/RestrictedActions` that the repo does not have yet.

## Known follow-ups

- Guest self-passport still shows no equipped items: `getUserProfile` returns `null` when catalyst + badges + photos all fail (always true for guests) before reaching the live-avatar path.
- Backlog *46: passport open blocks the whole panel on the equipped-items resolution (~3 sequential round trips). Needs a follow-up `userProfile` message so the panel can render with its own placeholders, like Unity.
- Backlog *47: hidden categories (Unity's `ComposeHiddenCategories`) are not filtered, only the body shape is. Needs per-item hides/replaces metadata carried through `resolveEquippedSet`.


robtfm:
Rebased onto current `main` (no conflicts) and added 11 commits on top of the original 20. Nothing in the passport feature itself changed; these are cleanups the review asked for, plus a few small unrelated fixes piled in so they don't need their own PRs.

## Cleanups on the passport work

- **One catalyst definitions resolver** — the emotes domain had a copy of wearables' resolve-by-urn (session cache + chunked lambda fetch) with only the query/response keys changed. Both now use `bridge-scene/src/domains/collections.ts`, parametrised by item kind with one cache per kind. Also hosts the single `thumbnailUrl` helper that replaces five hand-built `/lambdas/collections/contents/…/thumbnail` strings.
- **Shop deep links moved to `domains/marketplace.ts`** — they lived in the wearables domain and were imported from emotes, making emotes depend on wearables. Pure move.
- **Equipped-set resolver split by ownership** — `resolveEquippedSet` had grown an `indexTokens` flag whose "must be false for another user's urns" invariant lived in a comment and defaulted to the unsafe value. Now `resolveWearables` is the pure resolver for any address and `resolveEquippedSet` is the own-set wrapper that indexes tokens; the passport picks by `isSelf`.
- **Shared urn helpers (`domains/urns.ts`)** — `itemUrn` (tokenId strip) and `tokenUrnOf` (owned-catalog token urn) were duplicated across wearables and emotes. The emotes copy only handled collections-v2; the shared one also handles v1, which no emote uses.
- **One rarity palette** — `WearableCard` carried a per-rarity class palette, `lib/identity.ts` a second copy for inline styles, and the new `EquippedItemCard` used the latter while its sibling used the former (they already disagreed on legendary/mythic/unique). Now a single `[data-rarity]` block in `tokens.css` sets `--rm`/`--rl`; every tile (both cards, the backpack detail panel and emote slots) sets the attribute. The JS palette is deleted.
- **Passport tiles map to card props** — `EquippedRow` took `(Wearable | Emote)[]` and told them apart with `'slot' in it` / `'category' in it`. Each list now maps to `EquippedItemCard` props at the call site; `EquippedItemCardProps` is exported from the design index.

## Piled in

- **bridge-scene eslint baseline cleared** — `review.md` §0 lists it as a pre-push gate and it had 14 errors on `main` (one of them mine, a `return await`). All fixed via `eslint --fix`: method-signature style, braces around void arrow bodies, redundant boolean literal compares on non-optional booleans, `type` over `interface`, `async` on a promise-returning callback. No behaviour change.
- **`review.md` refreshed** — Node 24, bridge scene is type-checked by the SDK build (bare `tsc` trips tsconfig deprecations), lint is a gate, the 27 baselines that actually exist are listed, the stale "backpack-emotes renders empty" caveat is dropped (it renders ten tiles), and a note that viewport-sized captures pin the passport's wearables grid only, with the rest of the section covered by tier 1.
- **Stray `node_modules/.vite/vitest/…/results.json` removed** — force-added past the gitignore in #932; nothing reads it.

## Verification (on the final head)

- `npm run typecheck` clean; `npm test` 380/380; `npm run test:visual` 27/27 (the two refreshed passport/profile baselines eyeballed); bridge-scene `npm run build` type-checks clean and `npm run lint` reports 0 errors.
- Tested live against real profiles: other players' passports show their deployed wearables + emotes with SHOP on collectibles; own passport reflects the live avatar.

PR #1079, opened 2026-08-11
Stop scene code reaching the raw deno ops table

gonpombo8:
## Summary

Scene JS can reach `Deno.core.ops` — the full op table, minus every check the `~system/*` wrappers apply.

`op_require("~scene.js")` returns the deployed bundle's source and `require` evaluates it with `Deno.core.evalContext` in the runtime's own realm (`init.js`). There's no membrane — no `delete globalThis`, no `Object.freeze`, no `Proxy` — anywhere in `crates/dcl_deno/src/js` or `crates/dcl/src/js`. `EngineApi.js` proves the point: its first line is `const {…} = Deno.core.ops`, and it is loaded through the identical path as the scene bundle.

The SDK never exposes `Deno`, but that only constrains the creator's editor. What ships is a bundled `.js` that nothing re-derives from the TypeScript — a creator appends one line to `bin/index.js` after `sdk-commands build` and deploys to a world they own. The join event spawns the worker automatically; there is no review step anywhere in that path.

Two things reach through it.

**On the authoritative server — SSRF to cloud metadata.** `op_fetch_custom_client`'s `proxy` option re-targets the connection at an address the SSRF guard never sees: `assert_public_url` inspects the request URL, the proxy is what actually gets dialled. `proxy: { url: "http://169.254.169.254:80" }` reaches metadata behind a public-looking URL, and `redirect::Policy::none()` does nothing about it. `ca_certs` compounds it — a scene-supplied trust root for the server's own outbound TLS.

Separately, and needing no `Deno` at all: `assert_public_url` and the connect resolve the host independently, so DNS rebinding defeats the guard from plain SDK `fetch()`.

**Everywhere, including the desktop client — one line aborts the process.** The `deno_net` ops are registered (`js/mod.rs:54`), have no `~system/*` wrapper, and are reachable only this way. `NP::check_net` was `panic!()`, and `NP` wasn't even in the OpState so the lookup panicked before reaching it. Across the V8 boundary that's `panic_cannot_unwind` → SIGABRT, not a catchable JS error — verified by running it:

```js
Deno.core.ops.op_net_listen_tcp({hostname:"127.0.0.1",port:9999,transport:"tcp"}, false, false)
```

The `dcl_deno_ipc` sidecar hosts every scene in one process (`main.rs:124`), and losing it deliberately exits the engine (`lib.rs:178-184`). So any creator could deploy a scene that kills the explorer of everyone who walks into it — no permission prompt, no SDK involvement.

**What I checked and found sound.** Signed-fetch metadata is built in Rust from the scene's own `CrdtContext.hash` (`dcl/src/js/fetch.rs:73-86`), so a scene cannot claim to be a different scene, parcel or realm — and `~system/SignedFetch` already exports `getHeaders`, so the raw headers were never behind the wrapper anyway. Permission decisions live behind the `RpcCall` boundary in `crates/restricted_actions` (Web3 signing also requires user activity within the last second), so reaching an op directly doesn't dodge them. Super-user is gated on both paths. The `localStorage` ops are the address-prefixing overrides, so calling them raw changes nothing. I diffed every registered op against every op name referenced by the 15 system modules — `op_fetch_custom_client` and the `deno_net` ops are the only meaningful ones with no wrapper.

## The fix

1. **`op_fetch_custom_client` refuses scene-supplied transport in server mode** — `proxy`, `ca_certs`, `cert_chain`/`private_key` rejected outright; the scene gets a client with the same rules as the default one.
2. **Public-only DNS inside the server http client** — a `PublicOnlyResolver` enforces the rule at connect time, so the checked answer and the dialled answer are the same answer. Shared logic moved to `common::util::resolve_public_addrs`, which `assert_public_url` also calls.
3. **The deno permission structs deny instead of panicking**, and are put into the op state at runtime construction rather than at scene setup (a missing type is itself a panic, and this way no entry point can forget them). Raw socket access now raises an ordinary JS error.
4. **On the authoritative server only, the realm is sealed before the scene bundle runs** — `init.js` captures `Deno` privately and `require` passes it in as the `Deno` parameter of the module wrapper, so all 15 `~system/*` modules are unchanged and still see `Deno`; then `Deno`, `__bootstrap` and `__infra` are deleted from the global object. `__bootstrap.core` is the same ops table, so deleting only `Deno` would not have been enough. The deletion happens at that moment rather than at load time because the deno ext modules read those globals during their own evaluation, which happens first.

## Scope

1, 2 and 4 are gated on `is_server`. That's deliberate: the server is where scenes are multi-tenant and the host has credentials and a network position worth reaching, so the reward for escaping into the ops table is real there. On the desktop client the scene keeps seeing exactly what it sees today — same global, same wrapper parameter — so no deployed scene or preview can regress. The process-abort bug is fixed for everyone by 3, which needs no realm surgery.

The consequence worth stating plainly: on the client, `Deno.core.ops` stays reachable from scene code, so any op added in future is exposed there by default. Sealing the client realm is a follow-up decision, not something this PR forecloses — the mechanism is in place and it's a one-line change to the gate.

## What could break

The realm seal is the risky part, and it now only applies to the authoritative server. A server-side scene bundle legitimately touching `Deno` would stop working; I grepped `js-sdk-toolchain` and there are no references. The `~system/*` modules keep working via the injected wrapper parameter, covered by a test that requires them from scene code after the seal.

It assumes no ext module reads `globalThis.Deno` / `__bootstrap` lazily after startup (they import `ext:core/mod.js` at evaluation time) and that dynamic `import()` stays unavailable (no `module_loader` is configured). Both hold today; the first is the one to re-check on a deno bump.

One deliberate behaviour change outside the server gate: in preview, a hostname that *resolves* to loopback is now allowed, where before only a literally-loopback URL was.

## How to test

`cargo test -p dcl_deno` — 7 tests:

- `scene_cannot_reach_the_ops_table_on_the_server` runs a scene through the real `op_require` → `evalContext` path and asserts `Deno` / `__bootstrap` / `__infra` are gone, including via `Function("return this")()`.
- `client_realm_is_left_alone` asserts the opposite in client mode. The two are each other's negative control — the scoping is pinned, not incidental.
- `system_modules_still_work_after_sealing` requires `~system/EngineApi` and `~system/Runtime` from scene code, i.e. after the seal.
- `raw_socket_ops_raise_instead_of_aborting` calls `op_net_listen_tcp` from scene scope and expects a JS error. Before the change this test aborts the whole test binary with SIGABRT.
- three on the custom-client rules, one pinning that client mode still accepts a custom client with a proxy.

Manually: deploy a scene whose bundle calls `Deno.core.ops.op_fetch_custom_client({ ca_certs: [], proxy: { url: "http://169.254.169.254:80" } })`. Before, it builds a working client; after, `Deno` is `undefined` in scene scope on the server, and the op refuses the proxy if reached any other way.

## Not in this PR

- Sealing the client realm (see Scope).
- The WebSocket path (`op_ws_create`) still uses the pre-flight check only; pinning its resolution needs a different hook in `deno_websocket`.
- Blocking IMDS from the worker's network namespace on the multiplayer-server side — the defence-in-depth layer that makes a future regression here not instantly credential disclosure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3641,7 +3641,7 @@ dependencies = [
 [[package]]
 name = "deno_console"
 version = "0.168.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "deno_core",
 ]
@@ -3686,7 +3686,7 @@ checksum = "a13951ea98c0a4c372f162d669193b4c9d991512de9f2381dd161027f34b26b1"
 [[package]]
 name = "deno_fetch"
 version = "0.192.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "base64 0.21.7",
  "bytes",
@@ -3731,7 +3731,7 @@ dependencies = [
 [[package]]
 name = "deno_net"
 version = "0.160.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "deno_core",
  "deno_permissions",
@@ -3762,7 +3762,7 @@ dependencies = [
 [[package]]
 name = "deno_permissions"
 version = "0.28.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "deno_core",
  "deno_terminal",
@@ -3788,7 +3788,7 @@ dependencies = [
 [[package]]
 name = "deno_tls"
 version = "0.155.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "deno_core",
  "deno_native_certs",
@@ -3815,7 +3815,7 @@ dependencies = [
 [[package]]
 name = "deno_url"
 version = "0.168.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "deno_core",
  "urlpattern",
@@ -3824,7 +3824,7 @@ dependencies = [
 [[package]]
 name = "deno_web"
 version = "0.199.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "async-trait",
  "base64-simd 0.8.0",
@@ -3842,7 +3842,7 @@ dependencies = [
 [[package]]
 name = "deno_webidl"
 version = "0.168.0"
-source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#a476e471e54c289b59759327ae08c9fb9deee3b1"
+source = "git+https://github.com/robtfm/deno?branch=1_46_hotfix#4ecaf52e29460848d24b8300ec769e8b184dba47"
 dependencies = [
  "deno_core",
 ]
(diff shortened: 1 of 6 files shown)


robtfm:
Pushed a follow-up commit (`scene sandbox: egress policy for all scenes + close websocket DNS rebind`) and rebased onto current `main` — the #1069 stacking is gone now that it's merged. What's new on top of the seal / `NP` work:

### Egress: public-only for every non-preview scene (was server-only)
- The SSRF guard now runs for client fetch and WebSocket too, not just the authoritative server. A scene is untrusted code running with the user's network position, and the web build is already held to this by the browser's Private Network Access rules — so this brings the native client into line rather than tightening it past the browser.
- Gate is `!preview`, not `is_server`. `preview` widens the allowance to the local network (loopback + RFC1918 / IPv6 ULA) but **never** link-local / metadata — `169.254.169.254` stays blocked in every mode.
- Deleted the unguarded client-mode fetch client; both modes now go through a single guarded client with public-only DNS (`PublicOnlyResolver`). Custom-client proxies are vetted in client mode as well (a literal-IP private proxy is refused, since it skips the resolver).

### WebSocket DNS rebind closed at the connection
- A pre-flight `assert_public_url` alone can't pin a rebinding nameserver, because the connect re-resolves the hostname independently. Added `WebSocketPermissions::check_resolved` to the deno fork: `handshake_websocket` resolves the host **once**, the embedder vets the resolved addresses, and the connection dials exactly those — checked address == dialled address, with the hostname still used for TLS SNI. (This is the property the reqwest resolver already gave the fetch path.)
- Bumps the deno fork: `robtfm/deno` `1_46_hotfix` → `4ecaf52e`.

### `WebSocketPerms` moved into `create_runtime`
Now put alongside `TP` / `NP` at runtime construction, so every permission object exists from the moment the runtime does and no entry point can forget it.

### Realm seal kept
We enumerated the full scene-reachable op surface to check whether the seal could be dropped. For a non-super-user scene it's largely safe on its own — `deno_net` is neutered by `NP` (deny, not panic), `system_api` is registered only for super-user, and fetch / ws / storage ops are DCL-overridden — **but** the audit turned up one real gap the seal was masking: the async WebSocket send ops (`op_ws_send_binary_async` / `op_ws_send_text_async`, used by `WebSocketStream` and reachable only through the raw ops table) bypass the #1031 outbound-buffer cap. So the seal stays. That gap is fixed separately in #1081.

`op_fetch_response_upgrade` was also checked and is dead in this runtime — DCL replaced the fetch resource pipeline, so the deno `FetchResponseResource` it needs is never created and the op can only error.

PR #1090, opened 2026-08-12
Scope the super-user bridge BroadcastChannel per browser tab

### LINKED ISSUE #1089: with multiple browser clients, teleport actions from one window reach all windows
str:
- open 2 tabs at the same location
- `/goto somewhere` in one of them

expected: only 1 client goes to the new location
actual: both go to the new location

https://github.com/user-attachments/assets/294aec21-a586-4531-86c0-73487fd4333e

robtfm:
Fixes #1089

## Problem

`BroadcastChannel` is same-origin across **all** tabs, and the HUD ↔ bridge-scene transport uses the fixed name `bevy-ui-bridge` with no tab identity in the envelope. With two web clients open, a `/goto` typed in one tab is received and executed by every tab's bridge scene — and it's broader than teleports: any page→scene request (chat sends, emotes, settings, logout) fans out to all tabs, and scene→page streams (chat relay, player pose, profile) cross-pollinate every tab's HUD.

## Fix

A per-boot session id suffixes the channel name on both ends, **opt-in by the host page**:

- The page seeds `window.__bridgeSession` before booting the engine (react-web does this in `EngineHost`, and all page-side consumers open the channel via the new `bridgeChannelName()` — lazy `??=` init makes it load-order independent, including the e2e bridge spy's init script).
- `engine.js` forwards the id (if set) to the sandbox worker in the `INIT_WORKER` payload.
- `sandbox_worker.js` hands the super-user scene a `BroadcastChannel` wrapper that appends `#<id>` to every channel name — the bridge-scene code is untouched.
- No id seeded → names stay bare, exactly today's behavior.

Native is unaffected either way: the deno `BroadcastChannel` polyfill ignores channel names entirely (it's a per-process op-backed pipe), and each native instance runs its own CEF browser process, so nothing can cross.

## Why opt-in: creator-hub (cc @nicoecheza)

creator-hub's inspector embeds this engine in an iframe and talks to its editor-agent super-user scene over the same mechanism on a different channel (`dcl-editor-bus`) — but from the **parent document**, which can't see the engine iframe's `window`. An engine-generated id would have suffixed the scene side only and silently broken that integration on the next engine bump. With the id opt-in and unset, creator-hub keeps bare names and needs no changes.

It also gives creator-hub a migration path: multiple editor windows in one Electron session are same-origin, so `dcl-editor-bus` has the same cross-window leak this PR fixes for the web client. To adopt the fix there:

1. `engine-iframe.ts` generates a per-mount id and appends it to the engine URL (alongside `realm`/`position`/`systemScene`),
2. `host-boot.js` sets `window.__bridgeSession` from the query param before `__bevyLaunch` (the engine reads it at sandbox-spawn time, so anywhere before launch works),
3. the inspector-side bridges open `dcl-editor-bus#<id>`.

### DIFF
--- a/react-web/e2e/helpers.ts
+++ b/react-web/e2e/helpers.ts
@@ -64,7 +64,11 @@ export async function installBridgeSpy(page: Page): Promise<void> {
     if (w.__bridgeLog) return
     w.__bridgeLog = []
     try {
-      const ch = new BroadcastChannel(channel)
+      // Same lazy per-boot suffix as the app's bridgeChannelName() — the init script runs first,
+      // so it seeds __bridgeSession and the app's `??=` picks it up.
+      const ws = window as unknown as { __bridgeSession?: string }
+      ws.__bridgeSession ??= crypto.randomUUID().slice(0, 8)
+      const ch = new BroadcastChannel(`${channel}#${ws.__bridgeSession}`)
       ch.onmessage = (e: MessageEvent) => w.__bridgeLog!.push(e.data)
     } catch {
       /* BroadcastChannel unavailable — spy disabled */
--- a/react-web/src/engine/protocol.ts
+++ b/react-web/src/engine/protocol.ts
@@ -23,12 +23,12 @@ export const BRIDGE_CHANNEL = 'bevy-ui-bridge'
  */
 export function bridgeChannelName(): string {
   // Structural globalThis, not window/crypto: this module is also compiled into the bridge
-  // scene's program (type-only imports), whose tsconfig has no DOM lib. There crypto is absent
-  // and this falls back to the bare name — the scene never calls it anyway; its suffix is
-  // applied by the sandbox's injected BroadcastChannel wrapper.
-  const g = globalThis as { __bridgeSession?: string; crypto?: { randomUUID(): string } }
-  if (!g.crypto?.randomUUID) return BRIDGE_CHANNEL
-  g.__bridgeSession ??= g.crypto.randomUUID().slice(0, 8)
+  // scene's program (type-only imports), whose tsconfig has no DOM lib. (The scene never calls
+  // this at runtime — its suffix is applied by the sandbox's injected BroadcastChannel wrapper.)
+  const g = globalThis as { __bridgeSession?: string; crypto?: { randomUUID?(): string } }
+  // The id needs uniqueness, not unpredictability — Math.random covers insecure contexts
+  // (plain-http LAN testing), where crypto.randomUUID is absent.
+  g.__bridgeSession ??= g.crypto?.randomUUID?.().slice(0, 8) ?? Math.random().toString(36).slice(2, 10)
   return `${BRIDGE_CHANNEL}#${g.__bridgeSession}`
 }
 
--- a/react-web/src/engine/protocol.ts
+++ b/react-web/src/engine/protocol.ts
@@ -11,6 +11,27 @@ import type { SceneLoadingUi } from './generated'
 
 export const BRIDGE_CHANNEL = 'bevy-ui-bridge'
 
+/**
+ * The channel name to actually open: BRIDGE_CHANNEL plus a per-boot session suffix.
+ * BroadcastChannel is same-origin across ALL tabs, so the bare name lets one tab's HUD drive
+ * every tab's bridge scene (issue #1089). The id lives on `window.__bridgeSession` — lazily
+ * created by whichever page-side caller touches it first (this helper, or the e2e spy's init
+ * script). engine.js only FORWARDS it to the scene sandbox, which appends the same suffix
+ * (deploy/web/engine/sandbox_worker.js) — it never generates one, so embedders that share the
+ * bus across documents (creator-hub's inspector) keep bare names. EngineHost calls this before
+ * booting the engine, which is what opts this page in.
+ */
+export function bridgeChannelName(): string {
+  // Structural globalThis, not window/crypto: this module is also compiled into the bridge
+  // scene's program (type-only imports), whose tsconfig has no DOM lib. There crypto is absent
+  // and this falls back to the bare name — the scene never calls it anyway; its suffix is
+  // applied by the sandbox's injected BroadcastChannel wrapper.
+  const g = globalThis as { __bridgeSession?: string; crypto?: { randomUUID(): string } }
+  if (!g.crypto?.randomUUID) return BRIDGE_CHANNEL
+  g.__bridgeSession ??= g.crypto.randomUUID().slice(0, 8)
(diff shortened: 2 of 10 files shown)


react-web/src/features/engine/EngineHost.tsx
```diff
@@ -29,6 +30,11 @@ function injectEngine(): void {
   if (injected || window.__bevyBootConfig != null) return
   injected = true
 
+  // Seed window.__bridgeSession BEFORE the engine boots: engine.js forwards it to the scene
+  // sandbox, which then scopes the bridge BroadcastChannel to this tab (issue #1089). The suffix
+  // is opt-in per host page — engine.js never generates one — so this call is what turns it on.
+  bridgeChannelName()
```
regenesis-claw, line 34: [P2] The seed sits *after* the `if (injected || window.__bevyBootConfig != null) return` guard on line 30. In the pre-set boot-config path (a documented `boot.js` entry point — a host page sets `window.__bevyBootConfig` itself and this function bails), the engine boots with no session id, but the first page-side consumer to construct a channel afterwards (`EngineDriver`, `BridgeClient`, `installCefNativeBridge`) will still lazily seed one. The page then listens on `bevy-ui-bridge#<id>` while the scene is on the bare name, and the HUD goes silently dead — no error, just no messages.

Not reachable from this repo today (nothing else sets `__bevyBootConfig`), but hoisting the `bridgeChannelName()` call to just above the early-return guard makes the seed unconditional for this document and costs nothing.

react-web/src/engine/protocol.ts
```diff
@@ -11,6 +11,27 @@ import type { SceneLoadingUi } from './generated'
...
+ * bus across documents (creator-hub's inspector) keep bare names. EngineHost calls this before
+ * booting the engine, which is what opts this page in.
+ */
+export function bridgeChannelName(): string {
+  // Structural globalThis, not window/crypto: this module is also compiled into the bridge
+  // scene's program (type-only imports), whose tsconfig has no DOM lib. There crypto is absent
+  // and this falls back to the bare name — the scene never calls it anyway; its suffix is
+  // applied by the sandbox's injected BroadcastChannel wrapper.
+  const g = globalThis as { __bridgeSession?: string; crypto?: { randomUUID(): string } }
+  if (!g.crypto?.randomUUID) return BRIDGE_CHANNEL
```
regenesis-claw: [P2] `crypto.randomUUID` is only exposed in **secure contexts** — it's `undefined` on a plain `http://<lan-ip>` origin, which is a common way to test a web build on another device. There the helper silently returns the bare name, the id is never seeded, and #1089 is quietly back with no signal at all.

The fallback is consistent (page and scene both stay bare, so nothing breaks), but consider a non-crypto fallback so the scoping still applies, since the id only needs uniqueness, not unpredictability:

```ts
const rand = g.crypto?.randomUUID?.().slice(0, 8) ?? Math.random().toString(36).slice(2, 10)
```

Keeping the bare-name path for the no-DOM bridge-scene compile is still fine — that's what the `globalThis` structural cast buys you.

react-web/e2e/helpers.ts
```diff
@@ -64,7 +64,11 @@ export async function installBridgeSpy(page: Page): Promise<void> {
     if (w.__bridgeLog) return
     w.__bridgeLog = []
     try {
-      const ch = new BroadcastChannel(channel)
+      // Same lazy per-boot suffix as the app's bridgeChannelName() — the init script runs first,
+      // so it seeds __bridgeSession and the app's `??=` picks it up.
+      const ws = window as unknown as { __bridgeSession?: string }
+      ws.__bridgeSession ??= crypto.randomUUID().slice(0, 8)
+      const ch = new BroadcastChannel(`${channel}#${ws.__bridgeSession}`)
```
regenesis-claw, line 71: [P2] The `"${name}#${id}"` format and the id derivation are duplicated here from `bridgeChannelName()`. Since the init script can't import, the duplication is understandable — but if the separator or the slice length ever changes in `protocol.ts`, the spy just observes an empty channel and the tier-2 assertions fail as "nothing came over the bridge", which reads like a product bug rather than a harness drift.

Could the separator at least be passed in as an argument alongside `channel` (both are already serialized into the init script), so one constant governs both sides?

regenesis-claw (approved):
Approving — the mechanism is sound and correctly scoped.

Verified independently:
- Every page-side call site of the bare `BRIDGE_CHANNEL` constant is now routed through `bridgeChannelName()` (EngineDriver, BridgeClient, mockBridge, cefNativeBridge) — no missed opener left on the bare name anywhere in the repo, and no test/harness constructs a channel by literal name.
- The native claim in the description checks out: the deno polyfill in `crates/dcl_deno/.../init.js` stores `name` but routes everything through `op_bridge_to_page` / `op_get_bridge_stream`, so names are irrelevant there and CEF pages that suffix are unaffected.
- Opt-in default keeps creator-hub's cross-document `dcl-editor-bus` on bare names — no consumer break.
- Security: no new surface. The session id is a same-origin window global, and same-origin is already the trust boundary for BroadcastChannel; ordinary scenes still never receive the constructor. No secrets, injection, or auth changes.
- CI is fully green (Test Suite, Build and Deploy Web, Clippy, Rustfmt, Headless Smoke).

Reviewed against `react-web/review.md` §5. Findings are all P2 — nothing blocking.

**[P2] No tier-1 test for the new helper.** §5 asks for tests on changed behavior; `bridgeChannelName()` is cheap to pin: (a) two calls return the same name, (b) it seeds `window.__bridgeSession`, (c) it returns the bare name when `crypto.randomUUID` is absent. That last one is the branch most likely to silently regress.

**[P2] Commit trailer carries AI attribution** (`Co-Authored-By: Claude ...` plus the generated-with footer), which §5 explicitly disallows ("commit follows CLAUDE.md — imperative, no AI attribution"). Worth stripping before merge.

gonpombo8 (approved):

react-web/src/features/engine/EngineHost.tsx
robtfm: Done in 0a4e1fee — the seed now runs above the guard.

react-web/src/engine/protocol.ts
robtfm: Done in 0a4e1fee — falls back to `Math.random().toString(36).slice(2, 10)` when `crypto.randomUUID` is absent, so insecure contexts keep per-tab scoping.

react-web/e2e/helpers.ts
robtfm: Skipping this one: worst case is a harness-side drift failure in tier-2, and threading the separator through the init script adds plumbing without product value. If the format ever changes, the spy is a one-line follow-up.

Issue #605, opened 2026-02-17
foreign avatar transforms are not adjusted to scene base


robtfm:
`GlobalCrdtState`s int_sender currently only accepts `Vec<u8>` serialized crdt messages but could be modified to accept other things

we need to add more context to these messages to fix an outstanding bug - foreign avatar transforms are broadcast to all scenes (in `process_transport_updates`), but are not adjusted relative to the scene base translation

_Originally posted by @robtfm in https://github.com/decentraland/bevy-explorer/issues/603#issuecomment-3913974409_

PR #1093, opened 2026-08-13
fix: report scene-relative positions for foreign player transforms

gonpombo8:
## Summary

Foreign player `Transform`s were broadcast into scene CRDTs with raw world-space translations, parented to the `WORLD_ORIGIN` entity. Rendering resolved correctly through the parent chain, but any scene script reading `Transform.position` directly for a player entity (`getPlayer().position`, distance checks, trigger zones) saw world coordinates instead of scene-local ones — masked only when the scene sits at parcel `0,0`. On the headless authoritative server every connected player goes through the foreign-player path, so server-side game logic over player positions was wrong for any non-origin scene.

This is the bevy-side counterpart of decentraland/hammurabi-headless#27, which fixed the same issue (same `GlobalCenterOfCoordinates`/`WORLD_ORIGIN` convention) on the fork backend — with this change scenes observe the same player coordinates on either multiplayer backend.

Changes:
- Add `Localizer::Transform`: `WORLD_ORIGIN`-parented transform payloads get their translation offset by the scene origin and are re-parented to `ROOT`. This rides the existing per-scene localization infra from #653, so it applies to both the initial CRDT snapshot (`subscribe`) and live updates, across all execution paths (deno, deno-ipc, wasm, headless).
- `DclTransformAndParent` now registers that localizer instead of opting out via `PositionFree`.
- Drop the per-scene `WORLD_ORIGIN` entity seeding in `initialize_scene.rs` — no scene-visible payload references entity 5 anymore.
- Unit tests for the localizer (offset + reparent, scene-space pass-through, malformed payload pass-through).

Stacked on `feat/per-room-presence-isolation`.

## What could break

- Every scene now observes foreign players scene-relative and `ROOT`-parented. This is the SDK convention (and what the fork backend does after hammurabi-headless#27), but a scene that manually compensated for the old world-space behavior would now double-convert.
- The bridge scene's nametag distance math mixed `getPlayer().position` (world) with the camera transform (scene-relative); it was wrong for non-origin scenes before and is fixed, not broken, by this change.
- Scenes still receive a `DELETE_ENTITY`-free tombstone-less entity 5; anything that read the seeded `WORLD_ORIGIN` transform (none found in-repo, and it's not part of the SDK protocol) would now see no `Transform` on it.

## How to test

- `cargo test -p dcl_component` — covers the localizer directly.
- Manual: run a scene whose base parcel is not `0,0` (Genesis City scene or a world with a non-origin base parcel), connect a second client, and log `getPlayer({ userId }).position` for the remote player — values should be small scene-local coordinates matching the local player's convention, with parent `0` (ROOT) instead of `5`.
- Headless: spawn a non-origin scene through the multiplayer server, have a client join, and verify server-side `Transform` reads for the joined player are scene-local.

### DIFF
--- a/crates/dcl_component/src/lib.rs
+++ b/crates/dcl_component/src/lib.rs
@@ -32,6 +32,10 @@ pub enum Localizer {
     Unimplemented,
     /// Localize `PbAvatarMovementInfo`: offset `walk_target` (field 8) by scene origin.
     AvatarMovementInfo,
+    /// Localize `DclTransformAndParent`: world-space transforms (parented to `WORLD_ORIGIN`)
+    /// get their translation offset by scene origin and are re-parented to `ROOT`, so scenes
+    /// read scene-relative positions for foreign players.
+    Transform,
 }
 
 impl Localizer {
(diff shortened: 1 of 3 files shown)


robtfm:
fixes #605

happy with it, can't merge as it's part of a stack

robtfm (approved):

PR #1098, opened 2026-08-14
feat: read communities from the v2 address-only endpoints

LautaroPetaccio:
## Why

The v1 community reads are being deprecated. They exist only to embed profile data the service resolves per row — `registry.getProfiles(...)` behind the owner names, member rows and ban lists — and that profile fetching is being removed from the service. v2 is the same data minus those fields, so every client has to resolve identity itself before v1 can go; this is the explorer's side of that.

## What

Moves the bridge scene's communities reads onto the v2 endpoints and resolves the addresses they return through the **engine's own profile cache**, which needed a batch fetch to be usable for a list.

| Read | Before | After |
|---|---|---|
| browse list | `GET /v1/communities` | `GET /v2/communities` |
| members tab | `GET /v1/communities/:id/members` | `GET /v2/communities/:id/members` |
| announcements tab | `GET /v1/communities/:id/posts` | `GET /v2/communities/:id/posts` |

Create, join, leave and `/places` have no v2 counterpart, so they stay on v1 and the base URL now takes the version. Those four are writes and a profile-free read, so they don't block dropping the profile fan-out — but they do mean v1 can't be retired wholesale on this client.

Community ids are encoded into every path segment and query value they land in (and so is the member address on the leave path). They arrive from the service as uuids, so nothing changes on the wire; it just means an id can never become more than one segment. The events query uses `encodeURIComponent` rather than `URLSearchParams`, which the scene sandbox doesn't provide.

## The profile data v2 drops, and where it comes from now

v2 keeps the same envelopes and query params; the only difference is the profile fields:

- list — `ownerName` → `ownerAddress`
- members — `name` / `profilePictureUrl` / `hasClaimedName` → `memberAddress` (`friendshipStatus` is retained, so the "Add Friend" gate is unaffected)
- posts — `authorName` / `authorProfilePictureUrl` → `authorAddress`

The client already had a profile cache for exactly this — `ProfileCache`/`ProfileManager` in `crates/comms/src/profile/mod.rs`, shared by the nametags, chat, the passport and the friends panel. It just couldn't serve a list: `get_remote_profile` fetches one address per request, so a 100-member tab would have been 100 round trips. So this gives it a real batch rather than opening a second fetch path beside it:

- **`get_remote_profiles()`** asks for many ids in one request. The registry endpoint already accepts an id list — only the caller was single-id. Chunked, so a long list is a few requests rather than one per address. `get_remote_profile()` is now that same path with one id.
- **`ProfileManager::request_many()`** marks everything it doesn't already hold as `Pending` behind a single in-flight batch, so a caller can ask for a whole list and then read it back through `get_data` as usual. A profile that arrives from comms while the batch is in flight wins; anything the response omits is marked `Failed` rather than waited on forever.
- **`SystemApi::GetPlayerProfiles`** serves the scene from that cache, holding the request until every address has resolved or failed so the scene gets one answer instead of polling. It answers with the addresses it was **asked** about, never the `ethAddress` the deployed metadata claims (that field is written by whoever deployed the profile). It takes unique addresses only and caps a request at 1000: a pending request is rescanned every frame, and the real callers ask for ≤120 (100 members + 20 post authors), so the cap only fires on a caller bug — it drops the tail and logs. The op is super-user-only, so that bound is robustness against our own scene, not an untrusted-input boundary.
- **`fetchIdentities()`** in the bridge calls that instead of the catalyst. A member already on screen costs nothing, the realm is whatever the engine is connected to rather than a `peer.decentraland.org` guess, and there's one profile cache again instead of two.

Two behaviour notes:

- An address with no resolvable profile now renders as `0x1234…abcd` instead of blank. This is also the case v2 exists for on the list: it "never fails when the owner profile cannot be resolved", where v1 could fail the whole request.
- `chat.ts` is in the diff only because the bridge's remaining profile cache (the passport's, which also needs links/description/badges) is now keyed on a lowercased address in both its callers; without it the same wallet could be cached twice.

The page-facing protocol is untouched — the bridge still sends the same `Community` / `CommunityMember` / `CommunityPost` shapes — so the HUD, the mock bridge and the tier-1 tests needed no change.

## Testing

- `cargo clippy -p system_api_types -p system_bridge -p comms -p dcl -p dcl_deno --locked -- -D warnings` — clean
- `cargo check -p system_ui` — clean (the other `get_remote_profile` caller)
- `cargo fmt --all -- --check` — clean
- `npm run typecheck` (react-web) and bridge-scene `tsc --noEmit` — clean
- `npm test` — 326 passed / 49 files
- bridge-scene eslint — 16 errors, all pre-existing in files this PR doesn't touch (`outfits`, `pointer`, `proximity`, `session`, `wearables`, `world`)
- v1/v2 payloads diffed against production before writing the mapping; for the same rows the profile data matches what v1 embedded

Not run locally, both environmental:

- **`cargo clippy` over the whole workspace** — `ffmpeg-sys-next` and `cef-dll-sys` don't build on this machine (missing system ffmpeg/CEF). The crates this PR touches are covered by the per-crate run above; `dcl_wasm` is the one edit that isn't, as the `wasm32-unknown-unknown` target isn't installed here. Its wrapper mirrors the existing `op_get_permanent_permissions` / `op_set_bindings` pattern, but CI is the first real check on it.
- **Tier 1.5 visual** — this machine lacks the Playwright browser build the pinned version wants. It can't be affected either way: no React app source changed, and `?mock=1` renders from `mockBridge`, which never loads the bridge scene.

### DIFF
--- a/crates/comms/src/profile/mod.rs
+++ b/crates/comms/src/profile/mod.rs
@@ -5,7 +5,7 @@ use std::{io::Read, path::PathBuf, sync::Arc};
 use anyhow::anyhow;
 use bevy::{
     ecs::system::SystemParam,
-    platform::collections::{HashMap, HashSet},
+    platform::collections::HashMap,
     prelude::*,
     tasks::{IoTaskPool, Task},
 };
@@ -29,7 +29,7 @@ use super::{
 };
 use common::{
     profile::{LambdaProfiles, SerializedProfile},
-    rpc::{RpcEventSender, RpcResultSender},
+    rpc::RpcEventSender,
     sets::SceneSets,
     structs::PrimaryUser,
     util::{TaskCompat, TaskExt},
@@ -41,7 +41,6 @@ use dcl_component::{
     },
     SceneComponentId, SceneEntityId,
 };
-use system_bridge::{PlayerProfileData, SystemApi};
 use wallet::Wallet;
 
 pub struct UserProfilePlugin;
@@ -51,12 +50,7 @@ impl Plugin for UserProfilePlugin {
         app.add_systems(
             Update,
             (
-                (
-                    handle_player_profile_requests,
-                    drive_profile_fetches,
-                    request_missing_profiles,
-                )
-                    .chain(),
+                (drive_profile_fetches, request_missing_profiles).chain(),
                 process_profile_events,
             )
                 .before(process_transport_updates), // .in_set(TODO)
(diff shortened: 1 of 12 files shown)


robtfm:
Rebased onto `main` (now carrying #1117 and #1118) and refactored the engine side away. Your three commits are unchanged apart from the replay; everything below is one commit on top, so it reviews on its own. Previous head was `9a6c3e2f` if you want to diff against it.

### Why the engine half is gone

#1117 reworked the profile cache into per-address entries with a driver that owns all fetch policy — creating an entry is the whole of "queue a fetch", and `drive_profile_fetches` collects every due entry into as few registry requests as it can, batched per registry host. So the batching this branch built for itself (`Pending`, `batches`, `poll_batches`, `request_many`, `get_remote_profiles`) is now what the cache does by default, and a list consumer doesn't need a list API at all: ask about each address separately and the misses still collapse into one POST.

It turned out to be redundant twice over. `getPlayerData` already resolves an arbitrary address through that same cache — `op_get_player_data` → `RpcCall::GetUserData` → `ProfileManager::get_data`, with connected players short-circuited from their `ForeignPlayer` component ([`restricted_actions/src/lib.rs`](https://github.com/decentraland/bevy-explorer/blob/main/crates/restricted_actions/src/lib.rs#L1206-L1230)). `SystemApi::GetPlayerProfiles` was re-plumbing a path scenes have always had, so it and its four op wrappers, `PlayerProfileData` + its ts-rs export, and `handle_player_profile_requests` are all removed. The branch no longer touches Rust: 11 files became 3, +413/−100 became +124/−41.

### What the TS does now

`fetchIdentities` issues one `getPlayerData` per address behind `Promise.all`. The mapping follows the different payload — `displayName`, `avatar.snapshots.face256`, and `hasClaimedName`, which bevy serializes but the SDK's `UserData` typing doesn't declare, hence the local type widening.

`MAX_IDENTITIES = 200` replaces the op's engine-side cap, on the client side where the list actually arrives. Each address now costs a slot in the scene's per-tick RPC budget (1000, shared with every other call the bridge makes that tick), and the real callers ask for at most 120 — 100 members plus 20 post authors — so a service that ignored its `limit` param would truncate a list instead of exhausting the tick and silently failing unrelated calls alongside it. Anything past the ceiling still renders, as its shortened address.

Your `chat.ts` change is untouched, and `communities.ts` differs by one word in a header comment.

### Two behaviour differences worth knowing

- A bare-hash `face256` no longer resolves against `base_url` — the op's helper did that. Registry answers are absolute, so this only bites a catalyst-sourced profile, which then renders faceless rather than wrong.
- Your own address is served from `CurrentUserProfile` rather than the cache, and that path returns an error outright if the profile hasn't loaded. So an owner row that is *you*, opened before login settles, shows a shortened address instead of resolving a moment later.

### Related

#1118 was split out and merged first: `get_user_data` answered a request naming an already-connected player and then returned from the system instead of continuing the event loop, so the rest of that frame's requests went unread and both pending drains were skipped. One at a time that's invisible; ~120 at once with nearby players among them would have dropped rows. This branch needs it.

Verified on the rebase: bridge-scene `tsc --noEmit` and `eslint` clean on all three files. No Rust to check — `crates/` is byte-identical to `main`.

PR #1102, opened 2026-08-15
Isolate player presence per scene room via CRDT contexts

robtfm:
Isolates player presence per scene room on the orchestrated (multi-tenant) headless server, so a scene only ever sees the players who are actually in its room — replacing the room-slice approach explored in #1094.

## Design

`GlobalCrdtState` becomes a **Component**: one instance per scene room on the orchestrated server, and a single shared instance on the client. Isolation is topological — there is no mode flag anywhere.

- `CrdtContexts` indexes contexts by scene hash (`""` = the shared context). Scenes resolve via `for_scene_hash` — their room's context if one exists, else the shared one. The server's shared context has no transports, so a roomless scene fails closed.
- Transports bind to a context at connect time (`AdapterManager::connect(adapter, context)`; the context is threaded through the `Start*` events and `Transport.context`), so a packet can never reach another room's store.
- `ForeignPlayer.context` means the same wallet in N rooms is N entities, so per-room connect/disconnect is ordinary entity lifecycle.
- A per-context `PlayerIdAllocator` (free-list + generation bump on reuse) replaces the never-censused global allocator, which silently wrapped and aliased ids after ~400 cumulative joins.
- The `dcl_deno_ipc` sidecar tags global updates per context (`GlobalUpdate(presence_context, …)`) with a per-context receiver map, preserving the snapshot/replay continuity guarantee across scene re-mints.
- RPC scoping (`getConnectedPlayers`, `onPlayerConnected/Disconnected`) is by context equality; the message bus drops any packet whose declared scene id doesn't match its room's hash (forgery guard), inert on the client.

## Also included

- **comms: url authority for livekit scene-room connect** — `host()` dropped an explicit port, silently breaking adapter URLs like `ws://localhost:7880`.
- **comms: `DCL_SCENE_ROOM_ADAPTER` gatekeeper bypass** — a local/offline escape hatch that connects a supplied adapter string directly instead of minting one from the gatekeeper. Client-path only.
- **headless: finalize the guest wallet once** — `setup()` was re-finalizing the wallet `main()` had already finalized, which re-randomized the guest identity mid-startup.
- **presence-isolation test harness** (`local-harness/presence/`; `harness-util` is a workspace member so CI compiles it, but the matrix itself only runs locally) — an end-to-end matrix driving a livekit dev server, an orchestrated engine with two scenes in two rooms, and synthetic clients (client-mode headless): 28 assertions covering per-room presence isolation, foreign avatar transforms and scene-origin localization at off-origin parcels, getPlayersInScene on both server and client paths, onEnterScene/onLeaveScene (including a killed client), bus forgery, and adapterless-scene refusal.


gonpombo8:
## Review

The core mechanism is sound — context-as-component, transport→context binding at connect time, per-context id allocator, tagged IPC stream — and the client path is genuinely unchanged (one context ⇒ every `single()`/`for_scene_hash` resolves to the same entity). Findings cluster in the server path, mostly where "resolve the context" happens twice from different inputs.

---

### 1. The scene's context is resolved twice, from two different inputs — `initialize_scene.rs:700`

`load_scene_javascript` captures the broadcast receiver at `initialize_scene.rs:435` from `for_scene_hash(&definition.id)`, stashes it in `SceneLoading::Javascript`, and `initialize_scene` then labels that stream at `:700` with `for_scene_hash(&context.hash).to_bits()` — re-resolved, potentially many frames later. Two distinct problems fall out:

**a) Permanent bind to the wrong context (the durable case).** `for_scene_hash` falls back to the shared context when the room context doesn't exist yet, and there is no re-bind anywhere — nothing reacts to a context being spawned later. In `AddScene`, `orch.wanted.insert` / `portables.insert` run *before* the adapter branch, so the scene starts loading immediately while the room context only appears when the async gatekeeper mint lands (`drain_control_commands` → `connect`, `headless.rs:789`). If the scene wins that race it subscribes to the shared context and stays there forever. Same outcome on the `"adapter required outside preview mode"` path (`headless.rs:870`): `scene-failed` is emitted, but the scene has already been queued and loads anyway — with no room, permanently.

**b) Mismatched tag (the transient case).** If the mint lands *between* the two systems, the receiver is the shared context's but the tag is the room context's bits. `renderer_ipc_out` stores `global_receivers[room_bits] = shared_receiver` (`dcl_deno_ipc/src/lib.rs:243`), so the scene is fed the shared context's updates under the room's key — on a server the shared context has no transports bound, so it sees nobody. The reverse ordering (context despawned via `RemoveScene` in between) inserts the room's receiver under `shared_bits` and clobbers the shared entry for every other scene on that stream.

Fix for (b): capture the context entity alongside `global_updates` in `SceneLoading::Javascript` and thread that one value through, so the receiver and its tag can't disagree. (a) needs an actual answer — either hold the scene until its room context exists, or re-subscribe when one appears.

---

### 2. `get_players_in_scene` drops the context filter on the non-room-scoped branch — `restricted_actions/src/lib.rs:1314`

```rust
let room_scoped = scene_context != presence.contexts.shared();
… .filter(|(e, f)| {
    if room_scoped { f.context == scene_context }
    else { containing_scene.get(*e).contains(scene) }   // ← every context's players
})
```

The `else` branch iterates `others` — which is *all* `ForeignPlayer`s across every context — and filters positionally only. Any server scene that resolves to the shared context (case 1a above, which is durable) lands here and can list co-tenant rooms' players: exactly the leak this PR closes, and `get_connected_players` right above it doesn't have the escape hatch.

It's worse than positional filtering suggests, because on headless there is no position to filter on. `ForeignPlayer` entities spawn with `Transform::default()` (`global_crdt.rs:553`) and `foreign_dynamics` lives in `AvatarPlugin`, which `headless.rs:431` deliberately omits — so every foreign player sits at the origin and `containing_scene` collapses to "whoever owns parcel 0,0".

The two conditions want to be ANDed, not alternated: `f.context == scene_context && containing_scene.get(*e).contains(scene)`.

---

### 3. Cross-room forgery guard is missing on the `NonPlayer` arm — `global_crdt.rs:861`

The new guard is only on `NetworkUpdate::Player` (`global_crdt.rs:696`):

```rust
if let Some(room_hash) = &state.room {
    if scene.scene_id != *room_hash { continue; }
}
```

`NetworkUpdate::NonPlayer` at `:861` checks only `address != "authoritative-server"` and then calls `process_messagebus` with no room check at all. `string_senders` / `binary_senders` are process-wide `Local<HashMap<String, _>>` (`:481-482`) keyed by scene hash and shared across every context, so a message routes purely on its declared `scene_id`.

Reachability: `participant/plugin.rs:354` routes any participant whose identity isn't a valid h160 down the `NonPlayerUpdate` path. So a participant in room A identifying as `authoritative-server` posts a bus packet declaring scene B's hash and reaches scene B's bus in another room. Gated in practice by the gatekeeper controlling identities, but the asymmetry with the `Player` arm is the point — the same `if let Some(room_hash) = &state.room` guard belongs here.

---

### 4. Early return skips the participant despawn — `livekit/participant/plugin.rs:175`

```rust
if let Some(address) = participant.identity().as_str().as_h160() {
    let transport_id = *room_entity;
    let Some(sender) = transport_senders.get(transport_id) else {
        return;                       // ← skips everything below
    };
    …
}
…
commands.entity(entity).despawn();     // :214 — mandatory cleanup
```

`transport_senders.get` now resolves `Transport → transport.context → GlobalCrdtState`, so it can fail where nothing in this branch could fail before. When it does, the observer returns and the `LivekitParticipant` entity plus its `HostedBy` link and audio state leak permanently. Want `let sender = transport_senders.get(transport_id)` with only the send skipped. The identical pattern at `:332` and `:393` is fine — nothing follows there.

---

### 5. `DCL_SCENE_ROOM_ADAPTER` is honored in release builds, and bypasses the protocol allowlist — `comms/src/lib.rs:75`

```rust
.insert_resource(SceneRoomAdapterOverride(
    std::env::var("DCL_SCENE_ROOM_ADAPTER").ok(),
));
```

Unconditional in `CommsPlugin::build`, and `connect_scene_room:290` prefers it over the gatekeeper in every build including release. The env-var vector alone is arguable, but the protocol handling isn't: `connect_scene_room` passes the string straight to `AdapterManager::connect`, which accepts `signed-login:` and `ws-room:` (`lib.rs:400`, `:394`) — plus `fixed-adapter:` recursion. That's precisely the case the sibling path guards against, with a comment saying why:

> ONLY livekit adapters: any other protocol (signed-login, ws-room, fixed-adapter recursion) could make the engine sign a remote-chosen payload with its identity. — `headless.rs:780`

So the escape hatch can make a release client sign a handshake against an attacker-chosen endpoint with the user's wallet. At minimum it needs the same `starts_with("livekit:")` check; better, gate the whole resource on `preview.is_preview` the way `DisableSceneRoomGatekeeper` is gated.

---

### 6. Context guard ordered wrong in both archipelago systems — `archipelago.rs:99` and `:145`

At `:99` the `let Ok(context) = contexts.single() else { return; }` sits *before* `archi_events.read()`, so on failure the events are never drained and the `EventReader` cursor falls behind. At `:145` `manage_islands` despawns the previous island transport and only *then* `continue`s on the context check, leaving no island transport at all.

Both are unreachable today (single context on the client, archipelago disabled on servers), but note `single()` also fails on *multiple* matches — so these become live the moment anything spawns a second context on the client path. Cheap to just reorder.

---

### 8. The harness is a workspace member, so CI does build it — `Cargo.toml:78`

The PR body says `local-harness/presence/` is "not built by CI", but `members = ["crates/*", "local-harness/presence/harness-util"]` and `ci.yml` runs `cargo test --all --release --locked` on Linux/macOS/Windows plus `cargo fmt --all -- --check`. Both compile it.

`harness-util/Cargo.toml` declares `livekit` and `livekit-api` from a git branch as **non-optional and non-target-gated**, so every workspace build — including `--no-default-features` — now pulls livekit and its webrtc/native-tls build into the graph on all three runners. Either drop it from `members` and build it standalone via `--manifest-path`, or add it to `workspace.exclude`.

---

### Checked and fine

- `PlayerIdAllocator` free-list — no reachable double-`free`; `despawn_players` runs once per frame and `lookup.remove_by_right` is synchronous, so a same-frame re-join takes the recycled generation-bumped id.
- `next_global_update` — empty-map case is guarded with `future::pending()`, and `broadcast::Receiver::recv` is cancel-safe, so rebuilding the futures each `select!` iteration loses nothing.
- All `RpcCall` variant call sites updated; `Entity::PLACEHOLDER` in `agent_commands.rs` resolves safely to the shared context.
- `url.authority()` fix, `finalize_guest_wallet`, and the `load_scene_javascript` `continue`-retry (prior side effects are idempotent on retry).
- Suppressing `setup_primary_profile` under `server_mode()` is self-consistent: it's the only inserter of `UserProfile` on the local player, so `update_avatar_info`'s `None` branch never fires either — neither identity nor avatar data is written for `PLAYER`, which is the intent.

robtfm:
Thanks for the thorough review — everything is addressed in the ten commits now on the branch (2a6d8712..c8ae4987). Point by point:

**1 — double context resolution**: scenes now queue only after `connect` has registered their room context (returns whether the scene may load; the mint path carries the urn and queues on completion; the refusal paths queue nothing, so a room-less scene can never load). The resolved context is threaded through `SceneLoading::Javascript` so the ipc stream tag and the subscribed receiver can't disagree, contexts are reaped only once nothing references their hash (removed scenes still resolve during teardown), and `for_scene_hash` now panics on a miss on a multi-tenant server instead of silently falling back to shared. Relatedly, standalone `--server-mode` now requires `--preview` — production serving is orchestrated-only, so the shared-context single-scene mode is exclusively a local-dev flow.

**2 — get_players_in_scene**: fixed with the AND as you suggested. While verifying it we found the deeper problem behind your headless-positions observation: foreign players' engine-side transforms were *never written* on headless (the only writer lives in the render-bound avatar plugin set), which also left avatar colliders and trigger areas blind. The render-free `PlayerMovementPlugin` is now registered in headless so remote positions are real — with a server gate on the visual ground-snap path so co-tenant scene colliders aren't consulted (tenants physically overlap in world space).

**3 — NonPlayer guard**: added, mirroring the Player arm. (Reachability today is additionally blocked by livekit's duplicate-identity handling — a second `authoritative-server` can't coexist with the engine in a room — but the guard shouldn't depend on SFU behavior, and holds for whatever NonPlayer carries in future.)

**4 — participant despawn**: fixed; only the PlayerLeft notification is skipped when the transport is already gone. Agreed the `:332`/`:393` twins are fine as-is.

**5 — adapter override**: gated on preview mode and on the `livekit:` prefix (same rationale as the orchestrated gate), with a warn-once when ignored.

**6 — archipelago ordering**: both reordered.

**8 — harness-util workspace membership**: left as a member, deliberately — we tried the exclusion and it's worse than it looks, while the cost is smaller than it looks. The CI test job already builds the same livekit git stack for the explorer itself (default features include `livekit`), so harness-util's marginal cost there is the `native-tls` feature union; and the `--no-default-features` legs (wasm, headless smoke, clippy) never build harness-util at all. Exclusion outright fails to build: workspace `[patch]` tables don't apply to `workspace.exclude`d packages, so the bevy-fork pins would need duplicating into a second manifest (two places to update per fork bump), plus a separate target dir rebuilding the whole dependency tree for harness users. The PR body's "not built by CI" claim was wrong and has been corrected.

Beyond the findings, the same audit surfaced that `onEnterScene`/`onLeaveScene` were silently dead for orchestrated scenes (they host as portables, which `get_parcel` never resolves): room-scoped players now resolve scene membership by room context — a client only holds a scene room while that scene is its current scene, so room membership relays the client-side positional signal — while clients and standalone servers keep the positional path unchanged.

The local harness grew from 9 to 28 assertions to pin all of this down (each fix landed red-first where the path was reachable): foreign avatar transforms at exact spawn coordinates with scene-origin localization proven at an off-origin parcel, `getPlayersInScene` on both the server and client paths, enter/leave including a killed-client leave, and adapterless-scene refusal.

PR #1112, opened 2026-08-19
Support draco-compressed gltfs

### LINKED ISSUE #559: draco mesh compression
-126,97 and other scenes use draco mesh compression, bevy doesn't support it.

- native rust lib `draco-oxide` does not support decoding (as of 28/1/26)
- c++ bindings can't be compiled to wasm
but 
- there is a wasm lib that can be used in browser, potentially via the image_processing thread to decompress on disk / force reload
- this should work in deno too

if draco-oxide gets decoder support this would be much easier though.

robtfm:
Adds support for `KHR_draco_mesh_compression` gltfs by transcoding them to plain gltfs on first encounter, reusing the existing `image_processing` cache-rewrite flow.

## How it works

- Draco glbs fail bevy's loader at the `extensionsRequired` check. `check_assets` now also reads `AssetLoadFailedEvent<Gltf>` and, when the error names the draco extension, enqueues the cached bytes for processing (once per content hash per session; the processor independently verifies `extensionsUsed` against the actual bytes).
- A new draco pass in `process_gltf` (`processor/draco.rs`) runs before the BC7 texture pass: each compressed primitive's bufferView is decoded with [draco-oxide-decoder](https://github.com/reearth/draco-oxide) (pure rust, wasm32-clean, interop-tested against the reference C++ codec; pinned `=0.1.0-alpha.9` since alphas make no API-stability promises), decoded data is appended as new bufferViews, the existing accessors are pointed at them, and the extension is stripped. The JSON is otherwise untouched, and the result is persisted over the cached bytes — decode once, ever, per asset.
- On success the asset is reloaded in place (the original never loaded, so no image-swap is needed). In `gltf_container`, containers whose asset failed get a `GltfAssetFailed` marker; a small retry system re-runs marked containers through the normal load flow when their asset is loading or loaded again (a state only a repair-reload can produce).

Verified natively against gltf-pipeline-compressed models (textured + skinned): fail → transcode (~100ms) → reload → retry → spawned, including joints/weights and animator wiring; decoded positions refill the declared accessor bounds within quantization error.

## Tiny-texture passthrough

The transcoder previously BC7-compressed every embedded image, and the resize step rounds a 1-pixel dimension (or any aspect ratio over ~512:1) down to zero, failing the whole file. Harmless-but-wasteful in the normal flow; for draco content it would leave the asset permanently broken, and 1x1 textures are common. Each image is now gated with the same worth-compressing filter `check_assets` uses: tiny or undecodable images keep their original bytes, and the resize math is clamped so it can never produce a zero dimension. Covered by unit tests.

## Limitations

- Preview-server content uses `b64-` path-derived hashes, which are (correctly) excluded from the cache in both directions, so preview draco scenes are not repairable by this route — deployed content only. Preview support would need a session-scoped in-memory overlay in the ipfs reader; left as a follow-up.
- The wasm runtime path is verified end-to-end (textured, skinned, and live-scene draco content repair in the browser). Web testing surfaced an upstream decoder bug on wasm32, fixed via the pinned fork (see comments and reearth/draco-oxide#34).

fixes #559

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4124,8 +4124,7 @@ source = "git+https://github.com/robtfm/winit?branch=v0.30.x#a8c6812b563f9c0974a
 [[package]]
 name = "draco-nd-vector"
 version = "0.1.0-alpha.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4564672e9738d0f8050b9480ec1d1ccbde72ae7e55ce925ab20b9db896ebc30"
+source = "git+https://github.com/robtfm/draco-oxide?branch=fix%2Fwasm32-rans-refill#5acfcf680d4bd17222d910b74ff8bf76a8161575"
 dependencies = [
  "quote",
  "seq-macro",
@@ -4135,8 +4134,7 @@ dependencies = [
 [[package]]
 name = "draco-oxide-core"
 version = "0.1.0-alpha.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c0cee8775cb7c8bf283562163fee5ae1bc8448836774b615604ffce958028c5"
+source = "git+https://github.com/robtfm/draco-oxide?branch=fix%2Fwasm32-rans-refill#5acfcf680d4bd17222d910b74ff8bf76a8161575"
 dependencies = [
  "draco-nd-vector",
  "kiddo",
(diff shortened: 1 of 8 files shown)


robtfm:
Web testing surfaced a decoder bug: draco-oxide's rANS "branchless refill" fast path computed the state's msb with a 64-bit-only constant (`63 - leading_zeros()` on a `usize`), so on wasm32 it never refilled the entropy state. Any stream whose encoder happened to pick the branchless strategy decoded to garbage (distorted geometry) or desynced into framing errors ("unknown prediction scheme id"), while native decoded the same bytes correctly. Reproduced outside the engine with 12 real streams on wasm32; all decode byte-identically to native with the one-line fix.

Fixed upstream in reearth/draco-oxide#34; until that releases, `3e4d29db` pins the robtfm/draco-oxide fork carrying the fix via `[patch.crates-io]`.

Also confirmed while testing: the web asset-reprocessing flow (draco and BC7 alike) is inert against localhost-hosted realms by design — the service worker deliberately never caches localhost content so local scene edits stay visible. Deployed content is unaffected.

PR #1116, opened 2026-08-20
Fix remote profile resolution on .zone environments

gonpombo8:
## Summary

Authoritative servers on .zone realms could never resolve most players' profiles, so scenes saw truncated addresses instead of names (`AvatarBase` never populated). Three compounding causes in `get_remote_profile` / `ProfileManager`:

- The asset-bundle-registry host is hardcoded to **.org** (since #568), but the .zone and .org registries hold **separate profile namespaces** — the .org registry returns nothing (or an unrelated profile) for .zone players.
- A failed fetch was cached as `Failed` **forever**, silently — one timeout poisoned the address for the engine's whole uptime, even across rejoins.
- A stale registry answer (version older than announced) looped on refetch without ever falling back to asking the peer directly.

This PR picks the registry host from the peer's announced lambdas endpoint (realm fallback), retries failed fetches after 30s, logs fetch failures, and falls through to the comms `ProfileRequest` on stale answers.

## What could break

- Unresolvable addresses now retry every 30s instead of going quiet — slightly more registry traffic (one POST/30s per unresolved player, incl. guests until the comms fallback answers).
- Any environment whose lambdas endpoint doesn't contain `.decentraland.zone` still uses the .org registry (unchanged behavior).

## How to test

Namespace divergence repro (address from a live .zone session where the server logged a truncated name):

```bash
curl -s -X POST https://asset-bundle-registry.decentraland.org/profiles  -H 'content-type: application/json' -d '{"ids":["0x1d485820aea7ae6d090c5901fbaf69be224c1b76"]}'   # -> []
curl -s -X POST https://asset-bundle-registry.decentraland.zone/profiles -H 'content-type: application/json' -d '{"ids":["0x1d485820aea7ae6d090c5901fbaf69be224c1b76"]}'   # -> profile "vitZone"
```

End to end: run a headless authoritative server on a .zone realm, join with a desktop client, and watch the scene's `getPlayer()` name. Before this change it stays a truncated address forever; after, the real .zone name appears. For the retry fix: block `asset-bundle-registry.decentraland.zone` in /etc/hosts while the client joins, then unblock — before, the name never recovers; after, it resolves within ~30s (and the failure now shows as a `warn` in engine logs).

### DIFF
--- a/crates/comms/src/profile/mod.rs
+++ b/crates/comms/src/profile/mod.rs
@@ -72,8 +72,12 @@ impl Plugin for UserProfilePlugin {
 enum ProfileDisplayState {
     Loaded(Box<UserProfile>),
     Loading(Task<Result<Option<UserProfile>, anyhow::Error>>),
-    Failed,
+    Failed(web_time::Instant),
 }
+
+/// Failures are cached, so without a retry one transient error (timeout, dns, registry
+/// lag) would leave the address unresolvable for the engine's whole uptime.
+const PROFILE_FETCH_RETRY: std::time::Duration = std::time::Duration::from_secs(30);
 
 #[derive(Resource, Default)]
 pub struct ProfileCache(HashMap<Address, ProfileDisplayState>);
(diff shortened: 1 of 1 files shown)


gonpombo8:
Closed in favor of https://github.com/decentraland/bevy-explorer/pull/1116

PR #1119, opened 2026-08-20
feat: headless server viewer

pravusjif:
# Server viewer

A debug window on the authoritative scene server. Because the engine *is* a renderer, the
server draws its own ECS directly — there is no snapshot protocol and no second process, so
what you see cannot drift from what the server actually holds.

## Use it from a scene

```bash
DCL_SERVER_VIEWER=1 npm run start
```

`sdk-commands start` forwards the whole environment to the server it spawns, so this needs
no SDK flag and no change to your scene. A window titled **Decentraland Server Viewer**
opens next to the usual preview output, and the server logs `viewer=true`.

Controls: `WASD`/`Q`/`E` fly (`shift` faster, `ctrl` slower), right-drag look, wheel zoom
(orbit) or fly-speed (free), `Tab` cycle players, `F` frame everyone, `Esc` free camera.
Clicking a row in the player list focuses and orbit-follows that player. The camera starts
on the first player who connects — with nobody in the scene you are looking at the base
parcel, which on a large scene is nowhere near the action.

## What it shows

Scene geometry, plus one capsule per player at the position the server trusts — green for
the server's own placeholder player, blue for peers, yellow for the selection. Players are
capsules rather than avatars on purpose: the server knows a transform, a name and an
address, and nothing more.

Server mode trusts broadcast positions verbatim with no ground snap, so a capsule can sit
somewhere a client would never draw it. That divergence is the point.

## Caveats

- **Not a production-identical server.** The render plugins it enables are the client's, so
  memory and CPU differ from the measured headless figures. The engine prints a banner
  saying so at boot.
- Preview, single-scene only — refused with `--orchestrated`, and requires `--preview`.

Full details, including the plugin-ordering constraints: [`docs/headless-sdk-preview.md`](docs/headless-sdk-preview.md).

https://github.com/user-attachments/assets/4fc2effc-8cfe-4f1b-8825-591782f33932

### DIFF
--- a/deploy/headless/launcher/README.md
+++ b/deploy/headless/launcher/README.md
@@ -20,8 +20,33 @@ enabled; you rarely need to run it by hand.
 | `--production` | Production mode; disables preview-only behaviour. |
 | `--tick-hz <n>` | Scene tick rate. Defaults to 30. |
 | `--timeout <secs>` | Exit cleanly after N seconds. |
+| `--viewer` | Open a window rendering the server's own world. Preview, single-scene only. |
 
 `--scene-id`, `--private-key` and `--env` are accepted for hammurabi compatibility and ignored.
+
+## Debug viewer
+
+`--viewer` (or a truthy `DCL_SERVER_VIEWER`) opens a window showing what the server sees:
+the scene, remote players as capsules at the positions the server trusts, a clickable
+player list that focuses and orbit-follows a selection, and a free-fly camera.
+
+`sdk-commands start` forwards the whole environment to the server it spawns, so no SDK
+flag is needed:
+
+```bash
+DCL_SERVER_VIEWER=1 npm start
+```
+
+Controls: `WASD`/`Q`/`E` fly (`shift` faster, `ctrl` slower), right-drag to look, wheel to
+zoom, `Tab` to cycle players, `F` to frame everyone, `Esc` back to the free camera.
+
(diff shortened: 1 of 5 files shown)


robtfm:
i like the idea. but i think it would be nicer to have the viewer.rs code contained in a scene instead - the same approach we use for the system scene, editor, live scene viewer, movement controller, etc.

the idea would be to load a scene at startup from a name.dcl.eth source (with a cl arg override for dev), and then use the existing explorer apis to inspect the running scenes, show player overlays etc. https://github.com/robtfm/editor-scene would be a good starting point showing the scene inspection api, flycam etc. 

i'd also consider adding more of the normal client plugins, specifically the scene ui plugin, so that servers can render their own debug ui and have it visible in the viewer.

pravusjif:
I'm killing this initiative as it's getting super complex, and since we didn't get new reports of "players misplacement/impersonation" between different player entities since the bevy-headless migration, I don't think it's worth the trouble.

Following [Rob’s suggestion](https://github.com/decentraland/bevy-explorer/pull/1119#issuecomment-5368695345) to make the server-viewer its own scene would mean:
* (new repo) server-viewer scene (similar to the editor-scene precedent)
* bevy-explorer PR to expose the scene data for the server viewer (player entities data)
* refactoring this current PR 
* sdk-multiplayer-server repo PR (should be based on feat/bevy-storage-delegation) for the relay of the exposed scene data

PR #1141, opened 2026-08-27
feat: base domain

eordano:
`unity-explorer` recently gained the ability to define a 'base domain' that is not decentraland.{ENV} -- this closes that gap https://github.com/decentraland/unity-explorer/pull/9826

### DIFF
--- a/crates/collectibles/src/base_wearables.rs
+++ b/crates/collectibles/src/base_wearables.rs
@@ -307,8 +307,12 @@ pub fn default_bodyshape_instance() -> WearableInstance {
     .unwrap()
 }
 
-pub const CONTENT_URL: &str = "https://peer.decentraland.org/content/contents/";
-pub const BASE_URL: &str = "https://peer.decentraland.org/content";
+pub fn content_url() -> String {
+    common::base_domain::https_url_from_path("peer", "/content/contents/")
+}
+pub fn base_url() -> String {
+    common::base_domain::https_url_from_path("peer", "/content")
+}
 
 pub fn default_wearables(body_shape: &WearableUrn) -> impl Iterator<Item = WearableInstance> {
     match body_shape.as_str() {
--- a/crates/collectibles/src/base_wearables.rs
+++ b/crates/collectibles/src/base_wearables.rs
@@ -308,10 +308,10 @@ pub fn default_bodyshape_instance() -> WearableInstance {
 }
 
 pub fn content_url() -> String {
-    common::base_domain::https_url_from_path("peer", "/content/contents/")
+    common::base_domain::https("peer", "/content/contents/")
 }
 pub fn base_url() -> String {
-    common::base_domain::https_url_from_path("peer", "/content")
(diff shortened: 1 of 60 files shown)


robtfm:
Rebased onto main and reworked collaboratively with @robtfm — the branch now carries the original two commits plus a cleanup/extension stack. Summary of what changed:

**Core (native)**
- Fixed the headless compile error that was failing CI (`url` moved then reused).
- Naming: `https_url_from_path` → `base_domain::https(sub, path)`, joined by `wss(sub, path)` and `host(sub)`; dropped the empty-subdomain special case.
- `--base-domain` is latched before config load, and the CEF shell forwards it to the bundled HUD as `?baseDomain=` when non-default.

**Web threading**
- `?baseDomain=` query param does the same job on web (captured in `boot.js`, latched at the top of both `engine_init` and `engine_run`).
- Without the param, the **hosting origin** decides: a page served under `decentraland.zone` keys every backend to zone, `decentraland.org` (or anything else) to org. Same pure derivation implemented in the engine boot module and the HUD (`react-web/src/lib/baseDomain.ts`).
- All HUD data fetches (places, map, reels, lambdas, worlds-content-server, realm default) compose from the base domain via `serviceUrl()`. Outbound nav links and sso (origin-relative) deliberately untouched.

**Home scene rework**
- Config `server`/`location` → `home_realm`/`home_location`: `Option`s written ONLY by SetHomeScene. `None` = never pinned, so the home tracks the base-domain default; a pinned home persists as-is across base domains. Legacy keys are discarded (they were rewritten on every settings save, so none were intentional pins).
- `--server`/`--location` are now pure startup params (parity with web's `?realm=`/`?position=`) and are never merged into config.
- The HUD's post-login "Skip to Genesis Plaza" is now **Skip to Home**, targeting the pinned/derived home scene.

**Safety**
- Link-borne `?baseDomain=` outside decentraland.org/.zone renders the UntrustedLaunchGate interstitial (same treatment as `?systemScene=`) instead of booting; native mode is exempt.
- `?native=1` handling is compiled out of deployed web builds via a vite build-time define; it remains active in dev and the bundled native HUD.

**Fixes found along the way**
- Three engine-UI surfaces fetched `realm-provider.decentraland.org`, which no longer resolves at all: the change-realm `/realms` list (→ `realm-provider-ea`), the backpack collections lambda and the discover fallback image (→ `peer`, matching unity's canonical hosts).
- The sdk6 adaption layer stays hardcoded to org: `renderer-artifacts` is the one composed subdomain with no zone/custom deployment. Every other composed subdomain was probed and exists on both org and zone.

**Known limitations / follow-ups**
- The launcher would need to whitelist `--base-domain` before it can pass it through.
- The ethereum RPC follows the base domain but the path stays `/mainnet` — bevy has no contract table, so there's no Custom→sepolia arm like the unity PR.
- `builder-items` and the imposter CDN remain hardcoded.
- A fully self-hosted HUD needs `/auth` served at its own origin (sso is origin-relative).

PR #1156, opened 2026-08-28
Send `Pause` event on removing  `ShouldBePlaying`

### LINKED ISSUE #1148: av: stream players send no VsPaused when ShouldBePlaying is removed
Since #1053, stream-backed `VideoPlayer`/`AudioStream` entities emit no `VsPaused` video event when they stop being played. `should_be_playing_on_remove` (`crates/av/src/lib.rs:599`) only strips `ActiveReceiver`; pre-#1053 `stream_shouldnt_be_played` sent `VsPaused`.

`ShouldBePlaying` is removed on scene pause, `max_videos` eviction and the player going out of range, so a scene watching `VideoEvent` for a stream sees `VsPlaying` and then nothing until the entity is despawned (`VsNone`) or the stream resumes (`VsLoading`/`VsPlaying`).

Suggested fix: trigger `SetState { state: VsPaused }` for the container entity in `should_be_playing_on_remove` (needs the `ContainerEntity` lookup, which the other observers already do).

Verified with a scratch test on the merged code that the event list after `ShouldBePlaying` removal is empty.

hukasu:
Closes #1148

### DIFF
--- a/crates/av/src/lib.rs
+++ b/crates/av/src/lib.rs
@@ -606,7 +606,7 @@ fn should_be_playing_on_remove<T: AVPlayer>(
     commands.entity(entity).try_remove::<ActiveReceiver>();
 
     let Ok((has_stream, container_entity)) = av_players.get(entity) else {
-        error!("AVPlayer {entity} not a scene entity.");
+        debug!("AVPlayer {entity} not a scene entity.");
         return;
     };
     if has_stream {
--- a/crates/av/src/lib.rs
+++ b/crates/av/src/lib.rs
@@ -599,10 +599,23 @@ fn should_be_playing_on_add<T: AVPlayer>(
 fn should_be_playing_on_remove<T: AVPlayer>(
     trigger: Trigger<OnRemove, ShouldBePlaying<T>>,
     mut commands: Commands,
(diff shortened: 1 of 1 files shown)


crates/av/src/lib.rs
```diff
@@ -599,10 +599,23 @@ fn should_be_playing_on_add<T: AVPlayer>(
...
     trigger: Trigger<OnRemove, ShouldBePlaying<T>>,
     mut commands: Commands,
+    av_players: Query<(Has<Stream>, &ContainerEntity), With<T>>,
 ) {
     let entity = trigger.target();
     debug!("Stream {} no longer playing.", entity);
     commands.entity(entity).try_remove::<ActiveReceiver>();
+
+    let Ok((has_stream, container_entity)) = av_players.get(entity) else {
+        error!("AVPlayer {entity} not a scene entity.");
```
robtfm: triggers whenever a scene removes `VideoPlayer` (as `With<T>` is not true by the time `ShouldBePlaying<T>` gets removed).

suggest just removing the error! log, it only means `VsPaused` is not posted to an entity after the `T` is removed, which is not required anyway

robtfm (approved):

PR #1158, opened 2026-08-28
fix(pointer): max_distance is player distance; add max_camera_distance

robtfm:
## Problem

`PointerEvents.max_distance` was implemented as camera distance, following the protocol comment. Unity and godot measure it from the player and always have, so deployed scenes rely on that — in Genesis Plaza every pointer uses `maxDistance`, and while a scene virtual camera is active nothing is clickable in bevy without switching camera.

## Fix

Follow decentraland/protocol#470: `max_distance` is the player-distance threshold, `max_player_distance` is a deprecated alias (larger wins when both are set), and the new `max_camera_distance` takes over the camera-distance check. The combination rules keep their shape: only player → player check, only camera → camera check, both → OR, neither → player ≤ 10.

Vendored proto updated to match. The PROXIMITY interaction path is untouched (still `max_player_distance` only, 3 m default).

Verified against the wasm build with a test scene: a `maxDistance: 2` cube is clickable from a virtual camera 20 m away when the avatar is 1.5 m from it and not at 7.5 m; a `maxPlayerDistance: 2` cube behaves identically; a `maxCameraDistance: 5` cube (no `maxDistance`, avatar 16 m away) is clickable from a camera 4.5 m away and not from 8 m.

### DIFF
--- a/crates/dcl_component/src/proto/decentraland/sdk/components/pointer_events.proto
+++ b/crates/dcl_component/src/proto/decentraland/sdk/components/pointer_events.proto
@@ -25,33 +25,37 @@ option (common.ecs_component_id) = 1062;
 // --------------
 // PointerEvents can enforce interaction range using two independent distance checks:
 //
-// - Camera distance (`max_distance`): distance from the active camera to the target entity.
-// - Player distance (`max_player_distance`): distance from the avatar/player position to the target entity.
+// - Player distance (`max_distance`): distance from the avatar/player position to the target entity.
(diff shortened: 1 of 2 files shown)


robtfm:
Follow-up pushed: the PROXIMITY path now resolves the threshold from `max_distance` too (shared `max_player_distance()` helper, same alias rules as the cursor path); unit tests added for both. Entries that set only `maxDistance` on a proximity event previously fell back to 3 m.

PR #1211, opened 2026-09-02
Only move camera if player move is permitted

### LINKED ISSUE #1167: movePlayerTo: camera rotation applied even when the move is denied
`op_move_player_to` pushes `RpcCall::MoveCamera` in the same block as `RpcCall::MovePlayer` whenever `camera_target` is set (`crates/dcl/src/js/restricted_actions.rs`). The `move_camera` handler (`crates/restricted_actions/src/lib.rs:560`) only checks that the request comes from the containing scene — the MovePlayer permission is resolved later, in the move handler.

So when the user denies the MovePlayer permission prompt, the player stays put but the camera still snaps to face `camera_target`. The camera facing should be applied only once the move itself is approved.

hukasu:
Closes #1167

### DIFF
--- a/crates/common/src/rpc/mod.rs
+++ b/crates/common/src/rpc/mod.rs
@@ -128,6 +128,7 @@ pub enum RpcCall {
         to: Vec3,
         looking_at: Option<Vec3>,
         duration: Option<f32>,
+        camera_rotation: Option<Quat>,
         response: Option<RpcResultSender<bool>>,
     },
     WalkPlayer {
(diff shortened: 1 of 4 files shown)


robtfm (commented):
Looks good – the camera facing now only lands once the move is actually approved and in-bounds.

One small follow-up: the `RpcCall::MoveCamera` is now sent via `Commands`, and `move_camera` and `handle_player_move_requests` are unordered in the same `Update` set, so the command only flushes at the end of the schedule and the camera snaps one frame after the player is repositioned (previously both events were in the same batch and applied in the same frame when permission was already granted).

Adding `.after(handle_player_move_requests)` to `move_camera` in the plugin's `Update` set fixes that – bevy auto-inserts a sync point between explicitly ordered systems when the earlier one has deferred params, so the event is readable in the same frame.

PR #1212, opened 2026-09-02
DO NOT MERGE: sign auth-server requests with the ephemeral key

nachomazzara:
Scene wallet requests sent to auth-server (`remote_send_async` → `init_request`) now go out as signed-fetch requests: the ephemeral key signs `post:/requests:timestamp:metadata`, with the metadata carrying the keccak256 of the exact body bytes (`bodyhash`). The server can then tell a fresh request from a reused delegation. The login request keeps its current shape.

Depends on decentraland/auth-server#145 accepting the signature. Against the current server the extra headers are ignored, so this is inert until then. Written without a Rust toolchain at hand: not compiled, not formatted with rustfmt.

### DIFF
--- a/crates/restricted_actions/src/lib.rs
+++ b/crates/restricted_actions/src/lib.rs
@@ -1817,7 +1817,7 @@ pub fn handle_eth_async(
         tasks.push((
             response.clone(),
             IoTaskPool::get()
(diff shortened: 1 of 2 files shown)


robtfm:
closing since the linked pr was closed.

Issue #1172, opened 2026-08-31
rfc4 Reaction / ChatReaction: emoji reactions invisible on bevy


robtfm:
Protocol main adds `Reaction = 14` and `ChatReaction = 15` to the rfc4 `Packet` oneof. Not vendored; inbound packets with these tags currently decode to an empty body and are dropped (harmlessly, unlike the tag-13 case in #1165).

Unity ships full send+receive+render (`DCL/Chat/_Refactor/ChatReactions/Networking/MultiplayerReactionMessageBus.cs`), so unity users' emoji reactions are simply invisible to bevy users today. Receiving/rendering is the parity-critical half; sending needs HUD surface.

Issue #1165, opened 2026-08-31
rfc4 Packet tag 13: private SceneDrivenAnimation collides with upstream LookAtPosition


robtfm:
protocol main assigns `Packet.message` tag 13 to `LookAtPosition` (with `Reaction = 14`, `ChatReaction = 15`). Our vendored rfc4 proto uses tag 13 for the bevy-private `SceneDrivenAnimation`.

The two messages have incompatible wire types at field 1 (`float timestamp` vs `string scene_hash`), so decoding the other side's tag 13 is a prost `DecodeError` — which drops the **entire Packet**, not just the field (`Could not decode payload` in the livekit participant / websocket room receivers). This cuts both ways: a conforming client's LookAtPosition kills the whole packet on bevy, and our tag-13 SDA kills the whole packet on a conforming receiver.

Currently latent — unity has generated bindings for LookAtPosition but doesn't send or handle it. We should move first, before anyone ships it:
- move `SceneDrivenAnimation` to a high private tag (same collision-avoidance scheme as the nested `Movement` extensions at 10000/10001)
- vendor `LookAtPosition`/`Reaction`/`ChatReaction` at 13/14/15, even if unhandled, so upstream traffic decodes cleanly.

robtfm:
Resolved by #1215: the vendored rfc4 proto now carries LookAtPosition=13, Reaction=14, ChatReaction=15, so the tag collision is gone. Rendering those messages is tracked separately (#1172).

PR #1215, opened 2026-09-03
Remove conflict between `Packet` slot 13

hukasu:
Part of #1165 and #1172 without the implementation of `LookAtPosition`, `Reaction` and `ChatReaction`


robtfm:
will merge just before next org rollout, to avoid mismatched message ids between org and zone players

PR #1217, opened 2026-09-03
Nft shape color

### LINKED ISSUE #1180: NftShape: color ignored; NFT_NONE renders the Classic frame instead of frameless
PBNftShape `color` (background for transparent pixels) is never read, and `NFT_NONE` is mapped to the Classic frame glb rather than rendering frameless (`crates/nft/src/lib.rs:285-318`). The urn and the other 22 frame styles work.

Unity ships both: color applied in `NFTShape/Frames/MeshFrame.cs:49-55`, and NFT_NONE maps to a dedicated `NFTShapeLoader_None.prefab`.

hukasu:
Closes #1180

### DIFF
--- a/crates/nft/src/lib.rs
+++ b/crates/nft/src/lib.rs
@@ -1,7 +1,10 @@
 pub mod asset_source;
 mod extended_image_loader;
 
-use std::{f32::consts::FRAC_PI_2, path::PathBuf};
+use std::{
+    f32::consts::{FRAC_PI_2, PI},
+    path::PathBuf,
+};
 
 use asset_source::{Nft, NftLoader};
 use bevy::{
@@ -157,7 +160,7 @@ fn load_frame(
 
         // \o/
         let transform = if frame.0 == NftFrameType::NftClassic {
-            Transform::IDENTITY
+            Transform::from_rotation(Quat::from_rotation_x(PI))
         } else {
             Transform::from_rotation(Quat::from_rotation_x(-FRAC_PI_2))
         };
--- a/crates/nft/src/lib.rs
+++ b/crates/nft/src/lib.rs
@@ -296,6 +296,7 @@ fn load_nft(
                 MeshMaterial3d(materials.add(SceneMaterial {
                     base: StandardMaterial {
                         base_color_texture: Some(h_image.clone()),
+                        alpha_mode: AlphaMode::Blend,
                         ..Default::default()
                     },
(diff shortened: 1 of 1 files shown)


crates/nft/src/lib.rs
```diff
@@ -170,29 +202,41 @@ fn process_frame(
...
+            for spawned_ent in scene_spawner.iter_instance_entities(frame.instance) {
+                if let Some((mat, maybe_name)) = mat_nodes
                     .get(spawned_ent)
                     .ok()
-                    .and_then(|h_mat| mats.get(h_mat))
+                    .and_then(|(h_mat, name)| mats.get(h_mat).map(|mat| (mat, name)))
                 {
+                    let maybe_name = maybe_name.map(|name| dbg!(&name.0).to_lowercase());
+                    let mut mat_clone = mat.clone();
+                    if maybe_name.iter().any(|name| name.contains("frame")) {
```
robtfm: this catches an unintended `mat_frame` from classic; better would be "pictureframe" || "mat_ck"

crates/nft/src/lib.rs
```diff
@@ -170,29 +202,41 @@ fn process_frame(
...
-            for spawned_ent in scene_spawner.iter_instance_entities(frame.0) {
-                if let Some(mat) = mat_nodes
+            for spawned_ent in scene_spawner.iter_instance_entities(frame.instance) {
+                if let Some((mat, maybe_name)) = mat_nodes
                     .get(spawned_ent)
                     .ok()
-                    .and_then(|h_mat| mats.get(h_mat))
+                    .and_then(|(h_mat, name)| mats.get(h_mat).map(|mat| (mat, name)))
                 {
+                    let maybe_name = maybe_name.map(|name| dbg!(&name.0).to_lowercase());
```
robtfm:
```suggestion
                    let maybe_name = maybe_name.map(|name| name.0.to_lowercase());
```

robtfm (commented):
verified that this is the intended section for the colour - it's the background shown behind the image.

we also need to set the actual image to `AlphaMode::Mask(0.5)` to get the background to show though.

hukasu:
i made them `Blend` instead of `Mask`, is there anything that guarantees that NFTs alpha are only either on or off

robtfm:
no, blend works. unity uses mask(0.5) but blend seems better.

Issue #1173, opened 2026-08-31
Emote lifecycle: report AvatarEmoteCommand state (ES_FINISHED / ES_INTERRUPTED) and implement stopEmote


robtfm:
Protocol adds `state` (EmoteState: ES_STARTED/ES_FINISHED/ES_INTERRUPTED) to PBAvatarEmoteCommand, and a `StopEmote` RPC to RestrictedActions. We write only urn/loop/timestamp (`crates/avatar/src/animate.rs:445`), never lifecycle entries, and `stopEmote` has no export in `RestrictedActions.js` (calling it throws).

Unity ships both: `WriteAvatarEmoteCommandSystem.cs` writes all three states, and stopEmote is implemented (`RestrictedActions/GlobalWorldActions.cs`). `ES_INTERRUPTED` is also the protocol-blessed shape for the cancel-signal question discussed in #1153.

Note: the sibling `mask` field (AvatarMask, on the component and on the TriggerEmote/TriggerSceneEmote requests) is NOT written by unity either — masks can be vendored for wire-compat but implementing them is speculative; out of scope here.

PR #1222, opened 2026-09-04
headless: report remote players' emotes to server scenes

LautaroPetaccio:
## Why

Scenes read `AvatarEmoteCommand` on player entities to react to what players do (dance floors, emote-triggered doors, gesture mini-games). On the client that component is written by `avatar::animate` once the emote clip has loaded. The headless server omits the render-bound avatar plugin, so nothing ever wrote it there: a server scene could never observe a remote player's emote.

## What

A render-free `ForeignEmoteRelayPlugin` in `comms`, registered only by the headless binary, reports the full lifecycle the component can express:

- **Start** → `ES_STARTED` with the emote's real `loop`. Scene emotes carry it in the URN, the reference client's embedded animations come from a fixed table, and base/wearable emotes are looked up in their deployed metadata (`emoteDataADR74.loop`) on the catalyst, cached by pointer. A legacy bare name like `wave` resolves as a base emote. A start waiting on metadata holds everything queued behind it *for that player*, so a cancel can never overtake its start; other players are not blocked.
- **Stop** → `ES_FINISHED` when the server's one-shot timer expired, `ES_INTERRUPTED` when the player cancelled. Echoes the started emote's URN and loop so a scene can match it without bookkeeping. A stop with no start on record is dropped.
- **Mask** passes through from the wire. **Timestamps** are a host-side monotonic counter, which the SDK's grow-only set needs to sort and trim correctly.

Supporting changes:

- `crates/dcl_component`: the vendored `avatar_emote_command.proto` is brought up to the SDK's current version (adds `EmoteState` and the `AvatarMask` import); `PbAvatarEmoteCommand` is marked position-free so the global CRDT path accepts it.
- `crates/comms/src/pulse`: the decoder keeps the stop reason and mask it previously dropped; `PlayerMessage::Emote` carries both.
- `crates/comms/src/global_crdt.rs`: the transport handler emits a `ForeignEmoteEvent` per transition alongside the `EmoteCommand` it already maintained. The relay is three chained systems: queue events, flush what is resolvable, then bank landed lookups and request new pointers through the ipfs client against `peer.<domain>/content`. A failed lookup answers `false` for 60 s and is then retried; an unknown collection type reports a one-shot immediately rather than losing the event.
- `crates/avatar/src/animate.rs`: the client's own start report sets `state: ES_STARTED` explicitly. Client behaviour is otherwise unchanged; the plugin is deliberately not part of `GlobalCrdtPlugin` so the client never reports twice.

Bounds: 256-byte URN cap with whitespace/control-character rejection, 16 pending reports per player, 1024 cached pointers, 4 lookup batches in flight.

## Verification

- 9 new unit tests in `comms` (`foreign_emote_relay_tests`): start/finish ordering and monotonic timestamps, deferral on metadata with wire order preserved across a cancel, one-shot fallback for unresolvable identifiers, URN rejection and orphan stops, per-context routing, identifier classification, metadata parsing, failed-lookup retry window, cache eviction.
- `cargo test -p dcl_component` passes with the proto change.
- `cargo check --bin headless --no-default-features --features headless,livekit` passes.
- `cargo clippy -p comms -p dcl_component -p avatar --tests -- -D warnings` clean; `cargo fmt` applied.

Not yet exercised against a live Pulse server with real emoting peers; opening as a draft for that.

## Left open

- The client still reports only starts. Reporting finishes/interrupts there too needs changes inside the avatar animation system.
- Lookups target the catalyst rather than the realm's content server, which is where emotes are deployed today.

### DIFF
--- a/crates/avatar/src/animate.rs
+++ b/crates/avatar/src/animate.rs
@@ -31,7 +31,8 @@ use comms::{
 use console::DoAddConsoleCommand;
 use dcl::interface::CrdtType;
 use dcl_component::{
-    proto_components::sdk::components::PbAvatarEmoteCommand, SceneComponentId, SceneEntityId,
+    proto_components::sdk::components::{EmoteState, PbAvatarEmoteCommand},
+    SceneComponentId, SceneEntityId,
 };
 use ipfs::IpfsAssetServer;
 use scene_runner::{
@@ -451,6 +452,8 @@ fn animate(
                         emote_urn: broadcast_urn.to_string(),
                         r#loop: resolved_loop,
                         timestamp,
+                        mask: None,
+                        state: Some(EmoteState::EsStarted as i32),
                     },
                 );
             }
--- a/crates/comms/src/global_crdt.rs
+++ b/crates/comms/src/global_crdt.rs
@@ -29,13 +29,15 @@ use dcl::{
 use dcl_component::{
     proto_components::{
         kernel::comms::rfc4::{self, packet::Message},
-        sdk::components::{EmoteState, PbAvatarEmoteCommand, PbPlayerIdentityData},
+        sdk::components::{
+            common::AvatarMask, EmoteState, PbAvatarEmoteCommand, PbPlayerIdentityData,
+        },
     },
     transform_and_parent::{DclQuat, DclTransformAndParent, DclTranslation},
     DclReader, DclWriter, GlobalCrdtData, Localizer, SceneComponentId, SceneEntityId, SceneOrigin,
 };
 
-use ipfs::{ActiveEntitiesRequest, ActiveEntityTask, IpfsAssetServer};
+use ipfs::{ActiveEntitiesRequest, ActiveEntityTask, EntityDefinition, IpfsAssetServer};
 
 use crate::{profile::ProfileMetaCache, Transport};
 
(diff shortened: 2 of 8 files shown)


LautaroPetaccio:
Addressed the review in the second commit:

- **b64 canonicalisation (P1):** one `canonical_pointer` now covers both the pointer derived from a peer's URN and the pointers a catalyst answers with, so an answer always finds its request. Regression test `canonicalises_b64_segments_on_both_paths` drives a mixed-case `b64-` URN through the real answer-banking path (`bank_lookup`, factored out of the lookup system so tests can feed it `EntityDefinition`s without a network).
- **Bounded lookups (P1):** requests are chunked to 32 pointers and the wait queue is capped at 256. A start past the cap is reported as a one-shot at once, uncached, so no player's queue can stall on it. Test `caps_the_lookup_queue_and_chunks_requests` puts 300 players with distinct unresolved emotes through it: 256 queued, 44 reported immediately, chunks of 32 oldest-first.
- **Queue overflow (note):** a full player queue no longer drops anything. Only a start awaiting its `loop` can block a queue, so that head is force-reported as a one-shot and the queue drains; the old drop remains only as a last resort past twice the cap. Test `full_queue_force_resolves_its_blocked_head`.
- **Mask validation (note):** a relayed mask is written only if it names a value of the generated `AvatarMask`. Test `writes_only_masks_the_enum_defines`.

14 relay tests pass; clippy `-D warnings` on comms/dcl_component/avatar and the headless type-check are clean.

robtfm:
ref #1173 

appreciate the pr and the bug callout for headless.

i'll address both together, i think it's worth having a solution in one place for both the client and the server.

regarding the embedded animations, we need a base-scene-emotes collection, similar to base-emotes. as it stands those emotes fail in bevy client and server mode, a new collection would allow them to work everywhere.

LautaroPetaccio:
Hi @robtfm , shall I close the PR or will you continue the work from here?

robtfm:
Leave it as is please, I’ll go through properly and make sure I take the tests and behaviours correctly.

robtfm:
closing in favour of #1227. tests and behaviours kept (with the exception of correct loop reporting for unity embedded emotes)

Issue #1237, opened 2026-09-07
Manual paste fallback for the sign-in deep link


robtfm:
When the `decentraland://` handoff cannot reach the explorer (no handler on macOS without the launcher, a browser sandbox that blocks the launch, a broken registration), the user is stuck: the auth page infers failure from the absence of a window blur and shows "Could not open Decentraland" with a retry button, but never displays the link. Nothing from the OS surfaces it either, since the page launches the link from a hidden iframe precisely to suppress the browser and Launch Services alerts, and the identity id is only inside the link (the request id in the page url is client-minted, the server does not map it to the identity).

A manual fallback needs two halves:

- decentraland/auth: a "copy link" button on the `ContinueInApp` failure view
- bevy: a paste field on the waiting-for-sign-in dialog (builtin login and the react HUD, the latter needing a system-api message to hand the text over) that runs the pasted link through the same `authRequestId` matcher as the bridge-file poll, so a link for another login is still rejected

Issue #1236, opened 2026-09-07
Sign in on macOS without the Decentraland launcher installed


robtfm:
The deep-link sign-in (the flow that replaced `dcl_personal_sign`) needs something on the machine to receive `decentraland://` links. On macOS the explorer currently asks the user to install the Decentraland launcher, because only an app bundle can claim a scheme and the URL is delivered as an Apple Event rather than on argv.

Handling it ourselves would mean:

- add `decentraland` to `CFBundleURLTypes` in `deploy/macos/Info.plist` (the bundle already claims an unused `dclbevyexplorer` scheme)
- register a `GURL` Apple Event handler with `NSAppleEventManager` before winit's run loop starts (an objc2 `define_class!` with one selector; objc2 and objc2-foundation are already in the tree via winit), which writes a sign-in link to the launcher bridge file the same way the windows/linux handler process does
- deal with scheme competition: with both us and the launcher claiming `decentraland`, Launch Services picks whichever registered last, and the launcher never reclaims. The polite version calls the deprecated `LSSetDefaultHandlerForURLScheme` on every start, naming the launcher when it is installed and ourselves otherwise. That is fragile and is the main reason this was not done in #1238.
- cold-start place links arrive as an event after the boot location was chosen, so they need the teleport-in-place handling too

There is no working native macOS build to test any of this against right now.

The cleaner end state is our own scheme everywhere: a `deepLink` target-config entry for bevy in decentraland/auth pointing at `dclbevyexplorer://`, registered on every OS without competing with the launcher. That still needs the Apple Event handler on macOS, but removes the competition and the launcher dependency.

PR #1238, opened 2026-09-07
Sign in via the auth site's deep-link handoff

robtfm:
Native login has failed since auth-server 53db603a (2026-07-28) retired `dcl_personal_sign`: every attempt ended in "400: the dcl_personal_sign method is not allowed", on org and zone alike. The auth site now builds the AuthIdentity itself, POSTs it to `/identities` and hands the id back through a `decentraland://` link carrying the client's `authRequestId`; the client fetches the single-use, same-IP identity. This is the flow unity and the creator hub use.

## Receiving the link without taking the scheme

The launcher (the foundation client's installer) owns `decentraland://` wherever it is installed, and only ever registers at install time, so claiming the scheme from it would stick. Instead:

- the auth URL carries `bridgeOnly`, which makes the launcher write the link to its `deeplink-bridge.json` and exit without starting unity; the login polls that file for its own `authRequestId` and leaves other logins' links in place
- on windows/linux, when nothing handles the scheme, the explorer registers itself the way the launcher would be (HKCU `Software\Classes`, or a hidden `.desktop` plus `xdg-mime`), taking the raw link as its first argument. A registration of our own that differs from what we would write today is redone; a foreign one is never replaced, unless its exe (windows) or desktop file (linux) is gone
- on macOS only an app bundle can claim a scheme and the URL never arrives on argv, so the login asks for the launcher to be installed (#1236)

The registered explorer handles every link, not just sign-ins: a sign-in link is relayed to the bridge file before anything else starts; a place link's `realm`, `position` and `dclenv` fill the launch options (a hidden positional in `ClientOptions`, off `--help` and the web param table). That is also the argument shape the launcher passes to an explorer. Linux dev builds (and the tarball, run with `LD_LIBRARY_PATH=.`) find libcef through `LD_LIBRARY_PATH`, which a browser-spawned process lacks, so the registration bakes it into the Exec line with each entry made absolute; an AppImage registers its own path rather than the transient mount and bakes nothing, its libraries resolve by rpath.

## Behaviour matrix

| | foundation client (launcher) installed | not installed |
|---|---|---|
| **macOS** | Browser hands the link to the launcher, which writes the bridge file and exits (`bridgeOnly`); the login consumes it. Verified live. Place links keep going to the launcher, so to unity. | Login fails with "install the Decentraland launcher". Place links have no handler. #1236 |
| **Windows** | Same as macOS: the launcher's HKCR registration is detected and left alone. A leftover registration whose exe is gone (launcher uninstalled without cleanup) counts as free. | First login registers `"<exe>" "%1"` under HKCU. Each link spawns a fresh explorer: a sign-in link is written to the bridge file for the running one; a place link boots a new instance into its realm/position/dclenv. |
| **Linux** | n/a: the launcher ships dmg and nsis only. A foreign registration, if any, is respected; one whose desktop file is gone (stale `mimeapps.list`) counts as free. | Always this column. First login writes a hidden `.desktop` and sets it via `xdg-mime` (needs xdg-utils), with the current `LD_LIBRARY_PATH` baked in as absolute entries; an AppImage registers `$APPIMAGE` with no environment. Otherwise as windows. |
| **Web** | unchanged: the page supplies the identity | unchanged |

Not covered: a place link while an explorer is already running starts a second instance on windows/linux (the launcher would have written the bridge file instead); a running explorer polling the bridge file for teleports is #1239. A manual paste fallback for when no handoff can reach the explorer is #1237.

## Also in here

- the login verification code is gone (LoginNew's code channel returns None) and the AuthIdentity parser moves from system_ui into wallet
- the platform crate's timers standardise on `async_std::task::sleep`; `futures-timer` is dropped
- the windows/linux registration modules are CI-verified only; the macOS side and the bridge-file login were exercised on a mac with the launcher installed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8638,11 +8638,11 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-compat",
+ "async-std",
  "async-tungstenite",
  "bevy",
  "directories",
  "futures-lite",
- "futures-timer",
  "futures-util",
  "js-sys",
  "reqwest 0.12.15",
@@ -8651,6 +8651,7 @@ dependencies = [
  "spin 0.9.9",
  "tokio",
  "tungstenite 0.21.0",
+ "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "web-fs",
@@ -12448,6 +12449,7 @@ dependencies = [
  "reqwest 0.12.15",
  "serde",
  "serde_json",
+ "uuid 1.26.0",
  "web-time",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8656,6 +8656,7 @@ dependencies = [
  "wasm-bindgen-futures",
  "web-fs",
  "web-sys",
+ "winreg",
  "ws_stream_wasm",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -231,7 +231,6 @@ async-trait = "0.1.68"
 fastrand = "2"
 rand = "0.8.5"
 futures-util = "0.3.28"
-futures-timer = "3"
 async-native-tls = { version = "0.5", features = ["runtime-async-std"] }
 boimp = { git = "https://github.com/robtfm/boimp", branch = "feat/composite-bake-via-storage-texture" }
 crc = "3"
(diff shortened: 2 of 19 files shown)


hukasu (approved):
Linux OK

It only opens the login page on chrome though, so i had to include chrome on my env for it to work

PR #1245, opened 2026-09-08
fix: resolve `..` segments when mapping a content path onto the content map

dalkia:
**TL;DR** A glTF image uri like `../../textures/bark.png` never resolved: `normalize_path` only swapped backslashes, so the content-map lookup used `models/tree/../../textures/bark.png` and missed. Materials rendered with only their emissive map (Unity resolves the same uri fine). `normalize_path` now collapses `.`/`..`, and the `$content_file` parser plus `ContentMap` lookups go through it.

**Tests** — normalizer cases (`..`, `.`, backslashes, urls, climbing above root), `IpfsPath::new_from_path` on the exact path bevy_gltf produces, and a `ContentMap` lookup that climbs out of the gltf folder.

**Context** — found while building the Creator Hub model optimizer (decentraland/creator-hub#1566), which currently writes shared textures beside each model to work around this.

### DIFF
--- a/crates/ipfs/src/ipfs_path.rs
+++ b/crates/ipfs/src/ipfs_path.rs
@@ -295,7 +295,7 @@ where
                 file_path.push_str(stripped_file_name);
                 Ok(IpfsType::ContentFile {
                     content_hash,
-                    file_path,
+                    file_path: normalize_path(&file_path),
                 })
             }
             "$entity" => {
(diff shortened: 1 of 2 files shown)


dalkia:
Superseded by https://github.com/decentraland/bevy-explorer/commit/75464974a65d6afaca019530faaef68f44863b5b

PR #1301, opened 2026-09-18
Run Explorer in an opt-in OffscreenCanvas worker

eordano:
Runs Explorer's Bevy event loop and WebGPU rendering in a dedicated worker when built with `web-worker` and opened with `?renderWorker=1`. Pins the patched Bevy/winit revisions from robtfm/bevy#6 and robtfm/winit#1. Default browser startup and native builds remain available.

The page prepares and transfers the canvas using winit, then the worker attaches it before `engine_run`. Asynchronous bridges keep DOM state, console commands, loading status, scene-worker startup, WebAudio playback, and LiveKit/WebRTC on their appropriate threads. Browser startup no longer creates the unused native Kira voice manager.

Build the Wasm bundle with `--no-default-features --features livekit,social,web-worker`, regenerate the sandbox bundle as usual, and open the app with `renderWorker=1`. A build without worker support reports an explicit error if that URL switch is requested.


robtfm:
closing in favour of #1346

